PR #96, opened 2014-12-23
#1740 Genuine links for list tables

litenull:
I am not fond of this solution but I cannot think of another way to allow genuine link handling on tr/td elements.


litenull:
Thanks for the feedback @hooverlunch. I did evidently make a vague comment on a very specific ticket. I will implement the suggestion regarding color change on mouseover, review the ticket again and push the changes here.

I will also ping you more often if I have any questions.

Tomaz

PR #101, opened 2015-01-06
1302 making hyperlinks from urls in report content, index and show

brodavi:
added https://github.com/gregjacobs/Autolinker.js for auto-linking urls on report content.

@Tom, I noticed in the gulp build flow, browserify is used. However, we appear to be hard-including things in the repo like jquery.sparkline.js directly in the index.html.. for this issue, I have followed what appears to be standard procedure, but I think using browserify correctly (npm install modules, and require() them) will be better in the long run.


hooverlunch:
Looks good. I would agree with you re: browserify for future reference. Feel free to refactor as you go along.

PR #102, opened 2015-01-06
1287 feature clear filters

brodavi:
Added button to remove any filter applied (not including search string). Button hides if no filters are currently applied.


hooverlunch:
Lovely.

PR #103, opened 2015-01-06
Feature #1792 - Incident location autocompletion

mkuklis:


public/angular/sass/app.scss
```diff
@@ -257,3 +257,13 @@ li.collapse, li.collapsing {
...
+
+.pac-container {
+  z-index: 1051 !important;
+}
+
+.map-marker {
+  width: 20px;
+  height: 20px;
+  background-image: url('../images/map-marker.png');
+}
```
hooverlunch: Can we try to have newlines at ends of files. Just a nitpick.

mkuklis: Yes I will add a newline right now.

public/angular/templates/incidents/modal.html
```diff
@@ -40,19 +40,28 @@ <h4 class="modal-title">
...
           <div class="form-group">
-            <label for="latitude" class="col-lg-2 control-label">latitude</label>
+            <div>{{myForm.input.$error}}</div>
+            <label for="latitude" class="col-lg-2 control-label">Latitude</label>
             <div class="col-lg-9">
-              <input id="latitude" name="latitude" type="text" class="form-control" ng-model="incident.latitude">
+              <input id="latitude" ng-required="incident.longitude" min="-90" max="90" name="latitude" type="number" class="form-control" ng-model="incident.latitude">
+              <div class="has-error" ng-show="showErrors">
+                <div class="help-block" ng-show="form.latitude.$error.max || form.latitude.$error.min">Latitude is invalid.</div>
+                <div class="help-block" ng-show="form.latitude.$error.required">Latitude is required.</div>
```
hooverlunch: Can this say "Latitude is required if longitude is entered." Similar for longitude. 

Also if I enter 12.3 for longitude and 'xxx' for latitude, I get 'Latitude is required'. Should be 'Latitude is invalid'. Same happens for longitude.

hooverlunch:
This looks phenomenal. Great job. Just a few notes.

Also:

Can you make it so that if the location name field is blank but the lat/lng is not, the lat/lng gets displayed in the incident list location column. Otherwise it is blank with a place marker thing which looks weird. When doing so, please round the lat/lng values to 6 decimal places.

Can you please add lat and lng to the incident detail page.

Thanks! When these things are done you can go ahead and merge and close the issue.

mkuklis:
@hooverlunch thank you for reviewing this.

mkuklis:
@hooverlunch the changes should be now in place

PR #104, opened 2015-01-06
#1886 - adjusted incident modal

mkuklis:


hooverlunch:
Looks great. Just one small nitpick:

image

The lat/lng numbers should be rounded to 6 decimal places when being displayed, even if they are more than that in the DB. If someone tries to enter more than 6, it should be truncated on submit. The values returned from Google should also be truncated to 6 places.

Then please merge, all good.

PR #108, opened 2015-01-08
1869 logging enhancements

sapraaman:
1. Configurable request timeout. Server throws 500 in case of timeout. Client should be handling 500 and showing error.
2. Express api logging to show incoming requests
3. Detailed api request and response logging based on configurable flags
4. Removed console logging, fixed file logging
5. Separate logger for each process
6. Log socket.io events emitted to the browser clients

Notes:

1. Express logs incoming requests to the console. We get this functionality by single line of code. These logs come with colors etc.
We have also implemented our detailed request logging, but no colors there. Have couple of thoughts here, can discuss with you on skype.

2. Script to be fixed on the production server (to remove stderr.log, stdout.log and forever.log) when we deploy this on production. Its a minor change.


app.js
```diff
@@ -1,17 +1,31 @@
 process.title = 'aggie';
+
 var processManager = require('./lib/process-manager');
-var logger = require('./lib/master-logger');
+var initLogger = require('./lib/master-logger').init;
```
hooverlunch: Why does this need a `var`?

app.js
```diff
@@ -1,17 +1,31 @@
...
+var log = require('./lib/master-logger').log;
+
+// fork child at specific module path
+function _fork(modulePath) {
```
hooverlunch: Nice.

config/secrets.json.example
```diff
@@ -46,36 +46,52 @@
...
+      "log_requests": true,
+      
+      // log all api responses
+      "log_responses": true
+    },
+    "fetching": {
+      "filename": "fetching.log"
+    },
+    "analytics": {
+      "filename": "analytics.log"
```
hooverlunch: Lovely.

config/secrets.json.example
```diff
@@ -46,36 +46,52 @@
...
-      "silent": false
+    
+    "master": {
+      "filename": "master.log"
+    },
+    "api": {
+      "filename": "api.log",
+      
+      // api requests will be timed out if not responded
+      "request_timeout": 60,
```
hooverlunch: I don't think this belongs under the logger key.

config/server-prefs.json
```diff
@@ -1,3 +1,3 @@
 {
-  "fetching": false
+  "fetching": true
 }
```
hooverlunch: Newline pls.

lib/api.js
```diff
@@ -16,6 +19,62 @@ var express = require('express');
...
+    res.end(chunk, encoding);
+  };
+  
+  next();
+}
+
+// Log incoming api requests
+function logRequests(req, res, next) {
+  var requestLoggingEnabled = S(config.logger.api.log_requests).toBoolean();
+  var responseLoggingEnabled = S(config.logger.api.log_responses).toBoolean();
```
hooverlunch: Why is the S stuff necessary here? Isn't log_requests already a boolean?

lib/api/socket-handler.js
```diff
@@ -31,6 +32,15 @@ var SocketHandler = function(app, server, auth) {
...
   this.io.sockets.on('connection', function(socket) {
     var clientQuery;
+    
+    // override the emit function to add logging
+    var emit = socket.emit;
+    socket.emit = function(event, data) {
+      logger.debug(event, data);
+      
+      socket.emit = emit;
+      socket.emit(event, data);
```
hooverlunch: Could 41-42 just be `emit(event, data)`?

Why are you reassigning socket.emit on 41? Wouldn't that remove the overridden handler?

hooverlunch:
Looks good. Just a few notes and you can merge.

config/secrets.json.example
sapraaman: True, will fix

config/server-prefs.json
sapraaman: Noted

lib/api.js
sapraaman: to manage the scenario if the variable is skipped from configuration file

lib/api/socket-handler.js
sapraaman: On line 37, we store reference to original "emit", which is the actual meat and sends stuff back to browser.
On line 38, we create a new definition of "emit", where we log the data to be sent back.
On line 41, we reassign the original definition so that the original is called where ever needed from now on.
On line 42, we call the original to send data to the client

hooverlunch: I dunno, it still seems to me that since `emit` holds a reference to the original `socket.emit`, the first time the new `socket.emit` is called, it will reassign the original method to `socket.emit`, and the new `socket.emit` will get garbage collected since nothing will refer to it anymore. So the new `socket.emit` will only ever get called once per connection. If you replace lines 41-42 with just `emit(event, data)`, that won't happen.

PR #109, opened 2015-01-09
#1784 add user property in source model, load user property when source ...

sapraaman:
Add user property in source model, load user property when source is loaded

Note for Tom: Not related to this issue, but noting in general. Test cases are not using authentication at all. This stumped me for a long time. This should be looked into and test cases should use authentication, and code should fail if authentication is not provided.


lib/api/v1/source-controller.js
```diff
@@ -60,9 +74,12 @@ module.exports = function(app, user) {
...
   app.put('/api/v1/source/_events/:_id', user.can('edit sources'), function(req, res) {
     Source.resetUnreadErrorCount(req.params._id, function(err, source, numberAffected) {
-      if (err) res.send(err.status, err.message);
-      else if (!source) res.send(404);
-      else res.send(200, source);
+      if (err) return res.send(err.status, err.message);
+      else if (!source) return res.send(404);
+      Source.populate(source, [{path: 'user', select: 'username'}], function(err, source){
+        if (err) res.send(err.status, err.message);
+        else res.send(200, source);
```
hooverlunch: Not sure if it's necessary to populate here. Or even to return the source. Can you check if the client side code is actually using the returned source here?

lib/api/v1/source-controller.js
```diff
@@ -13,27 +13,39 @@ module.exports = function(app, user) {
...
+    
     Source.create(req.body, function(err, source) {
-      if (err) res.send(err.status, err.message);
-      else res.send(200, source);
+      if (err) return res.send(err.status, err.message);
+      
+      Source.populate(source, [{path: 'user', select: 'username'}], function(err, source){
+        if (err) res.send(err.status, err.message);
+        else res.send(200, source);
+      });
```
hooverlunch: Can we check if the client side code is using this returned source?

models/source.js
```diff
@@ -21,7 +21,8 @@ var sourceSchema = new mongoose.Schema({
   enabled: {type: Boolean, default: true},
   events: {type: Array, default: []},
   unreadErrorCount: {type: Number, default: 0},
-  lastReportDate: Date
+  lastReportDate: Date,
+  user: {type: mongoose.Schema.Types.ObjectId, ref: 'User', required: false},
```
hooverlunch: Isn't `required: false` the default?

Also: in other models the type for foreign keys like this is `String`. Should they eventually be refactored? What is the benefit of this style?

hooverlunch:
Nice work. Just a few notes.

lib/api/v1/source-controller.js
sapraaman: ok, will check and revert

lib/api/v1/source-controller.js
sapraaman: Will check

models/source.js
sapraaman: yes, required:false is default, just a habit.

ObjectId uses less space than string. Eventually, we should be moving to objectids.

hooverlunch:
Looks great. Please merge.

PR #110, opened 2015-01-09
#1763

sapraaman:
1. Removed _sourceNickname from report model. This can be populated from source id, which is already stored in report
2. Before saving report, nullify _incident if it is empty. This is needed because we use mongoose method to populate related document, and it has to be a valid value or null
3. Before saving, always set _statusWasModified and _incidentWasModified so that unnecessary events are not raised if nothing is modified
4. When _statusWasModified or _incidentWasModified is raised, load related source and incident and emit that to the client

Please note that with the above, the client should stop using report._sourceNickname and use report._source.nickname, report._source.type and report._incident.title

### DIFF
--- a/lib/api/streamer.js
+++ b/lib/api/streamer.js
@@ -114,15 +114,24 @@ Streamer.prototype._addReportUpdatesListeners = function(emitter) {
 
   // Listens to report updates
   emitter.on('report:status', function(report) {
-    Report.findById(report._id, function(err, report) {
-      if (err) self.emit('error', err);
-      else self.emit('reportStatusChanged', report);
-    });
+    Report.findById(report._id).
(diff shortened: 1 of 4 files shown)


hooverlunch:
Several notes here before I even look at the code.

Re your #1: the reason _sourceNickname was copied to Report is for performance reasons to avoid the need to lookup a source. Since a report's source never changes, there is no downside here that I can see. Mongo is all about denormalization. Please reinstate this unless you disagree, in which case let me know.

I would like this issue to make it such that _any_ changes to Report are streamed. This is because we are now adding other fields (read, flagged) to Report, and removing the status field. So please remove any reference to specific fields and just monitor and stream any changes to Report. This should be doable, no?

Re #4, if you put back _sourceNickname and _sourceType, you won't need to load the source.

PR #112, opened 2015-01-11
1747_user_profile_page, links to profiles

brodavi:
For the profile page, I added a placeholder for gravatar image, as I see it is almost implemented server-side. Tried to style as best I can, but I am definitely no designer.


models/incident.js
```diff
@@ -17,6 +17,7 @@ var schema = new mongoose.Schema({
   updatedAt: Date,
   storedAt: Date,
   assignedTo: String,
+  assignedToId: String,
```
hooverlunch: I believe this should be refactored into a full association. `assignedTo` should be removed and replaced with _assignedTo, and should use an ObjectId. Talk to Aman about this if you're not sure how.

public/angular/js/controllers/users/profile.js
```diff
@@ -0,0 +1,28 @@
...
+  '$scope',
+  '$rootScope',
+  '$stateParams',
+  'users',
+  'User',
+  function($scope, $rootScope, stateParams, users, User) {
+    $scope.users = users;
+
+    function init () {
+      if (!stateParams["userId"] || stateParams["userId"] == "") {
```
hooverlunch: Doesn't "" eval to false in JS? So second clause is not needed.

public/angular/js/controllers/users/profile.js
```diff
@@ -0,0 +1,28 @@
...
+        $scope.user = $rootScope.currentUser;
+      } else {
+        for (var x = 0; x < $scope.users.length; x++) {
+          if ($scope.users[x]._id === stateParams.userId) {
+            $scope.user = $scope.users[x];
+          }
+        }
```
hooverlunch: Please use an underscore method or other helper instead of writing this whole thing out.

hooverlunch:
Looking good. See a few notes above with the code. Also:

* In the users list, please remove 'View profile' button and instead make the username link to the profile. This is how the other tables will work so want to keep consistent.
* Thanks for taking the initiative on the gravatar thing, but can you remove it for now since it doesn't quite work.
* Can the role please be capitalized as it is in the user list.
* (Slightly unrelated) Can the title of the users page just be 'Users' and not 'Settings/Users'
* Then can the title of the profile page be Users / <username>, with 'Users' linking back to the user list.
* I like the icons, nice touch. Can you please use fa-group for the role, since a role is more like a group.
* In general, link colors should be blue always, not just on hover. Can you please fix this in the incident list, and make sure the link is blue in the users list.

brodavi:
Hi @hooverlunch  I addressed the comments you made here in the latest commit.

brodavi:
updated this old pull request to cleanly merge into develop

PR #114, opened 2015-01-11
Bug #1779 Username limitation on user creation

brodavi:
Username must be less than 15 chars, and only contain alphanumeric  characters or the underscore.

### DIFF
--- a/public/angular/js/controllers/users/form_modal.js
+++ b/public/angular/js/controllers/users/form_modal.js
@@ -86,7 +86,14 @@ angular.module('Aggie')
 
     $scope.save = function(form) {
       if (form.$invalid) {
-        $scope.showErrors = true;
+        // username errors (required gets triggered if aggieusername is invalid?)
+        if (form.username.$error.aggieusername) {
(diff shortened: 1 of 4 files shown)


public/angular/js/controllers/users/form_modal.js
```diff
@@ -86,7 +86,14 @@ angular.module('Aggie')
 
     $scope.save = function(form) {
       if (form.$invalid) {
-        $scope.showErrors = true;
+        // username errors (required gets triggered if aggieusername is invalid?)
```
hooverlunch: Why is there a question mark in this comment? Are we not sure how this works?

hooverlunch:
This looks good. For the username length, if you look at the incident modal, there is a nice interaction technique there for incident name where it lets you know how many chars you have left and won't let you type more than the allowed amount. Could we do it like that for consistency's sake? Everything else looks good. Thanks.

brodavi:
merging and closing branch

ticket http://redmine.sassafras.coop/issues/1779

PR #115, opened 2015-01-11
Bug #1773 Incident map should have maximum zoom level

brodavi:


public/angular/js/services/map.js
```diff
@@ -42,13 +42,17 @@ angular.module('Aggie')
       google.maps.event.addListener(m, 'click', function() { self.show_info_window(this); });
     });
 
-    if (self.markers.length == 0)
+    if (self.markers.length == 0) {
       self.map.setCenter(new google.maps.LatLng(0, 0));
-
-    // else use bounds determined above
-    else
+    } else if (self.markers.length === 1) {
```
hooverlunch: This won't necessarily work, because what if there are two markers close enough to each other to end up with a very high zoom? Couldn't you do fitBounds as below, but then check to see what the zoom ended up as, and if it's too high, reduce it?

brodavi: That's a much better idea :)  Okay, doing fitBounds, adjusting if zoom is too high. There was some weirdness with fitBounds (it is an async fn), but found fix on stackoverflow, seems to work great.

hooverlunch:
Looks great. I just quickly fixed the indent. Good to merge.

PR #117, opened 2015-01-12
Feature #1735 New settings page, no more fetching streaming

hooverlunch:


lib/api/v1/fetching-controller.js
```diff
@@ -5,10 +5,10 @@ var app = express();
 var auth = require('../authentication')(app);
 var user = require('../authorization')(app, auth);
 var logger = require('../../logger');
+var prefs = require('../../../config/server-prefs.json');
```
hooverlunch: Did you check with Aman about splitting this out? I know he had some thoughts on the matter.

lib/api/v1/fetching-controller.js
```diff
@@ -25,20 +25,15 @@ app.put('/api/v1/fetching/:op', user.can('toggle fetching'), function(req, res)
...
-    app.emit(signal);
-    
-    return res.send(200);
-  });
+  // save fetching status to server-prefs.json
+  prefs["fetching"] = fetching;
+
+  // tell BotMaster to start or stop fetching
+  var signal = fetching ? 'fetching:start' : 'fetching:stop';
+  app.emit(signal);
```
hooverlunch: Could we make this one line?

public/angular/js/controllers/fetching.js
```diff
@@ -2,60 +2,23 @@ angular.module('Aggie')
...
     };
 
-    var parseStatus = function(status) {
-      if (typeof status == 'string') {
-        return status === 'true';
-      } else {
-        return !!status;
-      }
+    $scope.save = function () {
+      Fetching.set($scope.on, null, failure);
```
hooverlunch: Setting fetching from the new settings screen doesn't seem to be working.

public/angular/sass/app.scss
```diff
@@ -50,6 +50,10 @@ a {
   }
 }
 
+.navbar-nav > li > ul {
+  background-color: darken($navbar-green, 10%);
+}
```
hooverlunch: The color looks good. Can we make the menu items a bit roomier? Perhaps same font size and height as in nav bar?

public/angular/templates/settings.html
```diff
@@ -1,10 +1,39 @@
...
+    <li><a ui-sref="users">Users</a></li>
+  </ul>
+</div>
+
+<div ng-controller="FetchingController">
+  <table class="table">
+    <tbody>
+      <tr>
+        <td>
+          <h5>Fetching</h5>
```
hooverlunch: The table is presently too wide I think. Can you make it substantially narrower. Also usually the left column is narrower than the right in tables like this. Can we see how that looks?

public/angular/templates/settings.html
```diff
@@ -1,10 +1,39 @@
...
+               data-toggle="buttons">
+            <label class="btn btn-info"
+                   ng-click="toggle(true)"
+                   ng-class="{active:on===true}">
+              <input type="radio"> On
+            </label>
+            <label class="btn btn-info"
+                   ng-click="toggle(false)"
+                   ng-class="{active:on===false}">
+              <input type="radio"> Off
```
hooverlunch: Can we go with a flatter style and more contrast for on/off? Take a look here:

image

This is from http://bootswatch.com/superhero/, which is our bootstrap theme.

hooverlunch:
Hi David. Good job so far. Multiple comments inline.

brodavi:
@hooverlunch okay I tried implementing all comments. Also, reverted direct-access to server-prefs.json as per @sapraaman 's advice. However, @sapraaman , the tool you are using, 'nconf' is acting weird for me. Setting fetching on/off works correctly, but the response for /api/v1/fetching/:op is always false, resulting in the toggle switch always reflecting 'false', even when fetching is actually on. Can you take a look please?

@hooverlunch one of the items in the redmine issue http://redmine.sassafras.coop/issues/1735  is: "Updating settings should be done using /settings endpoint instead of specialized /fetching endpoint. Please refactor." Do you mean for me or @sapraaman to refactor away the /fetching endpoint?

hooverlunch:
It's up to you guys who does it. /fetching just needs to go away. Thanks.

brodavi:
@hooverlunch okay, factored out /fetching into more general /services endpoint. Since the fetching controller really wasn't doing anything but settings, I essentially renamed the fetching controller to the settings controller and made the appropriate event emitter changes.

Note for Reference: For anyone anywhere searching for nconf bugs, if you have a configuration pulled from two different files, you are going to run into confusion. Don't do it. Configurations should be unique.

hooverlunch:
Looking better. I pushed some small style tweaks. There are two major outstanding issues:

1. For me, if you set fetching to 'ON', click 'Save', and refresh the page, it shows as 'OFF', even though fetching is really on.
1. The issue description requested to eliminate socket streaming of fetching status. This wasn't done. fetching:start should still be emitted in the controller, but it should not be sent out on sockets. All mention of fetchingStatusUpdate (including tests) should be removed. 

In future, please ask me if something isn't clear in an issue description.

No need to resubmit this again. Just please fix the above two issues, ensure the tests are passing, and then you can merge and close.

PR #118, opened 2015-01-12
1895 report table changes

mkuklis:


public/angular/js/controllers/incidents/report_incident_modal.js
```diff
@@ -24,6 +24,7 @@ angular.module('Aggie')
       });
 
       modalInstance.result.then(function(report) {
+        report.read = true;
```
hooverlunch: Will the `read` also get set in the DB?

hooverlunch:
Code looks great.

A few notes

* I'm getting a 'false' in the date column:

image

* Read reports should have a gray bg as in mockup
* One bug that I'd love if you could fix (probably not your doing): if you pick an incident when there is a 'N new reports' blue bar at the top, it loads in those reports as soon as you pick the incident. It should only do that when you click the blue bar.
* The 'x' is a little low. Could you move it up to the corner:

image

* Please ensure incident is not set to no-wrap:

image

* The flag column should be a fixed width, of similar width as in mockup.
* Flag should also be a little bigger.
* Can you please remove extra space on right side of media icon:

image

* Please remove bg color change on mouseover for all table rows. I think it's misleading now that the whole row is not clickable.
* Instead, can you try making the content link a block, with min-height equivalent to the min height of a row, so that the whole cell is clickable, like the repository listing here: https://github.com/hooverlunch

mkuklis:
@hooverlunch the additional changes should be in place for: http://redmine.sassafras.coop/issues/1895

hooverlunch:
Looks great. Thanks.

PR #119, opened 2015-01-12
1892 new incident structure

mkuklis:
I introduced db migrations based on https://github.com/tj/node-migrate

in order to run outstanding migrations you can run:

````bash
migrate up
`````


hooverlunch:
`migrate up` suggests that node-migrate should be installed with -g? If so this should go in install instructions?

brodavi:
End users have different use-cases, but I almost never use -g. Installing local versions of supposed-to-be-global modules takes up more space and can be annoying to run. But sometimes you have a billion projects, and not all of them can use the same version of the same supposed-to-be-global module.

This is my own personal opinion on the subject, so take it with a grain of salt, but I would like to see the node community start putting instructions in README.md explaining clearly when someone should or shouldn't use -g

PR #121, opened 2015-01-13
Feature #1300 Footer with version number

hooverlunch:


public/angular/sass/app.scss
```diff
@@ -267,3 +272,22 @@ li.collapse, li.collapsing {
...
+  position: absolute;
+  background-color: #009446;
+  height: 20px;
+  width: 100%;
+  bottom: 0px;
+  left: 0px;
+  .version {
+    margin-top: 5px;
+    position: absolute;
+    right: 10px;
```
hooverlunch: Can the bg of the footer be a light gray color instead of the heavy green. Also can we get a bit more vertical padding both within the footer, and between the main content and the footer. You can see here how it's tight:

image

PR #125, opened 2015-01-20
1744 filter reports by author

brodavi:
Filtering reports by author http://redmine.sassafras.coop/issues/1744


brodavi:
gah, not master

PR #127, opened 2015-01-21
Feature 1896 Report detail screen changes

brodavi:
http://redmine.sassafras.coop/issues/1896

Using mongo population to fill out _incident on reports. Pulling list of incidents then mapping the association was awkward, I think. We still need to 'resolve' incidents and others (for listing when assigning), but I think this approach is more intuitive.

### DIFF
--- a/public/angular/js/controllers/incidents/report_incident_modal.js
+++ b/public/angular/js/controllers/incidents/report_incident_modal.js
@@ -19,7 +19,6 @@ angular.module('Aggie')
         size: 'lg',
         controller: 'IncidentSelectModalInstanceController',
         templateUrl: '/templates/incidents/report_incident_modal.html',
-// !!!!! NOTE DAVID SCOPE IS HERE BECAUSE NEED PARENT? I FORGET
         scope: $scope,
         resolve: {
           incidents: ['Incident', function(Incident) {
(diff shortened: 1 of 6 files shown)


brodavi:
not sure if this pr is relevant anymore

hooverlunch:
Generally these changes look good but this branch will need to be rebased against the current develop. I tried to merge but couldn't figure it out.

One note: I'm not sure why `report` got changed to `r` in a bunch of places. Can you please revert this. IMO using single letter variable names is only appropriate in a very small scope, e.g. an iterator function or similar.

public/angular/js/controllers/incidents/report_incident_modal.js
```diff
@@ -25,9 +25,15 @@ angular.module('Aggie')
...
 
       modalInstance.result.then(function(report) {
-        report.read = true;     
+        report.read = true;
         Report.update({id: report._id}, report, function(response) {
-          $scope.$parent.r = report;
+          // we're getting the report here because it is easier than
+          // getting the incidents list. we have just updated
+          // _incident, which is an ObjectId. now we want the populated
+          // object again
```
hooverlunch: This comment doesn't make total sense to me. Please consider revising. What is the purpose here? What is $scope.$parent? Why are we setting its report value? Michal also did some work on streaming updates to report so this stuff may not be necessary.

brodavi:
okay @hooverlunch please take a look and see if this latest rebase/touchup is kosher. thanks!

the reason for the 'pulling the updated report' thing, which I replaced with a local lookup for speed, is that the population of the incident object does not happen on report update.

PR #129, opened 2015-01-26
1900 batch mode

mkuklis:
http://redmine.sassafras.coop/issues/1900

### DIFF
--- a/lib/api/v1/report-controller.js
+++ b/lib/api/v1/report-controller.js
@@ -66,6 +66,12 @@ module.exports = function(app, user) {
     });
   });
 
+  app.get('/api/v1/batch/:_id', user.can('edit reports'), function(req, res) {
+    Report.checkoutBatch(req.params._id, function(err, reports) {
+      if (err) res.send(err.status, err.message);
+      else res.send(200, reports);
+    });
+  });
   return app;
 };
 
--- a/lib/api/v1/report-controller.js
+++ b/lib/api/v1/report-controller.js
@@ -66,12 +66,33 @@ module.exports = function(app, user) {
     });
   });
 
-  app.get('/api/v1/batch/:_id', user.can('edit reports'), function(req, res) {
+  // Grab new batch to given user
+  app.get('/api/v1/report/batch/:_id', user.can('edit reports'), function(req, res) {
     Report.checkoutBatch(req.params._id, function(err, reports) {
       if (err) res.send(err.status, err.message);
       else res.send(200, reports);
     });
   });
+
+  // Mark selected reports as read
+  app.post('/api/v1/report/batch/_read', user.can('edit reports'), function(req, res) {
+    if (!req.body.ids || !req.body.ids.length) return res.send(200);
+    
+    Report.find({'_id': {$in: [ req.body.ids ]}}, function(err, reports) {
+      if (err) return res.send(err.status, err.message);
+      if (reports.length === 0) return res.send(200);
+      var remaining = reports.length;
+      reports.forEach(function(report) {
+        // Mark each report as read only to catch it in model
+        report.read = true;
+        report.save(function(err) {
+          if (err) return res.send(err.status, err.message);
+          if (--remaining === 0) return res.send(200);
+        });
+      });
+    });
+  });
+
   return app;
 };
 
--- a/lib/api/v1/report-controller.js
+++ b/lib/api/v1/report-controller.js
@@ -66,25 +66,52 @@ module.exports = function(app, user) {
     });
   });
 
+  // Cancel batch
+  app.put('/api/v1/report/batch/cancel', user.can('edit reports'), function(req, res) {
+    Report.cancelBatch(req.body.id, function(err) {
+      if (err) res.send(err.status, err.message);
+      else res.send(200);
+    });
+  });
+
   // Grab new batch to given user
   app.get('/api/v1/report/batch/:_id', user.can('edit reports'), function(req, res) {
     Report.checkoutBatch(req.params._id, function(err, reports) {
       if (err) res.send(err.status, err.message);
-      else res.send(200, reports);
+      else res.send(200, {results: reports, total: reports.length});
     });
   });
 
   // Mark selected reports as read
-  app.post('/api/v1/report/batch/_read', user.can('edit reports'), function(req, res) {
+  app.post('/api/v1/report/_read', user.can('edit reports'), function(req, res) {
     if (!req.body.ids || !req.body.ids.length) return res.send(200);
-    
-    Report.find({'_id': {$in: [ req.body.ids ]}}, function(err, reports) {
+
+    Report.find({_id: {$in: req.body.ids}}, function(err, reports) {
       if (err) return res.send(err.status, err.message);
       if (reports.length === 0) return res.send(200);
       var remaining = reports.length;
       reports.forEach(function(report) {
         // Mark each report as read only to catch it in model
-        report.read = true;
+        report.markAsRead();       
+        report.save(function(err) {
+          if (err) return res.send(err.status, err.message);
+          if (--remaining === 0) return res.send(200);
+        });
+      });
+    });
+  });
+
+  // Flag selected reports
+  app.post('/api/v1/report/_flag', user.can('edit reports'), function(req, res) {
+    if (!req.body.ids || !req.body.ids.length) return res.send(200);
+
+    Report.find({_id: {$in: req.body.ids}}, function(err, reports) {
+      if (err) return res.send(err.status, err.message);
+      if (reports.length === 0) return res.send(200);
+      var remaining = reports.length;
+      reports.forEach(function(report) {
+        // Mark each report as flagged to catch it in model
+        report.flag();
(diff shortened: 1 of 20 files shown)


lib/api/v1/report-controller.js
```diff
@@ -66,6 +66,60 @@ module.exports = function(app, user) {
...
+      reports.forEach(function(report) {
+        // Mark each report as flagged to catch it in model
+        report.flag();
+        report.save(function(err) {
+          if (err) return res.send(err.status, err.message);
+          if (--remaining === 0) return res.send(200);
+        });
+      });
+    });
+  });
```
hooverlunch: Seems like it should be possible to do this in one db query? Or will that mess with the event stuff?

models/report.js
```diff
@@ -63,6 +70,17 @@ schema.post('save', function() {
...
 
+schema.methods.flag = function() {
+  this.flagged = true;
+  this.markAsRead();
+};
+
+schema.methods.markAsRead = function() {
+  this.read = true;
+  this.checkedOutBy = null;
+  this.checkedOutAt = null;
```
hooverlunch: This seems like an issue. What if I click a report and go to detail page. That should mark as read but not erase checkout.

models/report.js
```diff
@@ -129,4 +147,65 @@ Report.findSortedPage = function(filter, page, callback) {
...
+    this.lockBatch.bind(this, userId, ITEMS_PER_BATCH),
+    this.loadBatch.bind(this, userId, ITEMS_PER_BATCH)
+  ], function(err, results) {
+    if (err) return callback(err);
+    callback(null, results[2]);
+  });
+}
+
+// release items in older batches
+Report.releaseBatch = function(callback) {
```
hooverlunch: releaseOldBatches

models/report.js
```diff
@@ -129,4 +147,65 @@ Report.findSortedPage = function(filter, page, callback) {
   });
 };
 
+Report.checkoutBatch = function(userId, callback) {
+  async.series([ 
+    this.releaseBatch, 
+    this.lockBatch.bind(this, userId, ITEMS_PER_BATCH),
+    this.loadBatch.bind(this, userId, ITEMS_PER_BATCH)
```
hooverlunch: Why twice?

hooverlunch: Why not just have lockBatch return the batch?

models/report.js
```diff
@@ -129,4 +147,65 @@ Report.findSortedPage = function(filter, page, callback) {
...
+// load a batch for user
+Report.loadBatch = function(userId, limit, callback) {
+  var conditions = {
+    checkedOutAt: { $ne: null }, 
+    checkedOutBy: userId
+  };
+
+  Report.find(conditions).limit(limit).exec(callback);
+}
+
```
hooverlunch: With all these functions called ___Batch it's feeling like the batch functionality should be refactored into a separate class. No need for a new model, just a POJO. What do you think?

public/angular/js/services/report.js
```diff
@@ -6,6 +6,10 @@ angular.module('Aggie')
   return $resource('/api/v1/report/:id', null, {
     'query': { isArray: false },
     'save': { method: 'PUT' },
-    'update': { method: 'PUT' }
+    'update': { method: 'PUT' },
+    'markAsRead': { method: 'POST', url: '/api/v1/report/_read', isArray: false },
+    'flag': { method: 'POST', url: '/api/v1/report/_flag', isArray: false },
```
hooverlunch: These should both be PATCH or PUT I think.

public/angular/js/services/report.js
```diff
@@ -6,6 +6,10 @@ angular.module('Aggie')
...
+    'grabBatch': { method: 'GET', url: '/api/v1/report/batch/:id' },
```
hooverlunch: This should also technically be PATCH since the database is being changed.

public/angular/sass/app.scss
```diff
@@ -43,6 +56,14 @@ a {
...
 }
 
+.margin-left {
+  margin-left: 10px;
+}
+
+.margin-bottom-lg {
+  margin-bottom: 20px;
+}
+
```
hooverlunch: A lot of the class names here are un-semantic (they don't describe the content of the element, just the style). I always was taught this was not such a good practice, but maybe they're doing things differently nowadays?

public/angular/templates/reports/batch.html
```diff
@@ -0,0 +1,41 @@
...
+  <div class="pull-left margin-left padding-top-bottom">
+    <button class="btn btn-default" ng-click="markSelectedAsRead()">Mark as Read</a>
+  </div>
+  <div class="pull-left margin-right padding">
+    <button class="btn btn-default" ng-click="flagSelected()">Flag</a>
+  </div>
+  <div class="clearfix"></div>
+</div>
+
+<p ng-hide="pagination.total > 0 ">Sorry, but we couldn't find any reports assigned to you.</p>
```
hooverlunch: This is good, but if there are no reports available to check out when the user clicks 'Grab Batch', can there be a modal telling them so, rather than sending them in here?

lib/api/v1/report-controller.js
mkuklis: @hooverlunch yes you are right it would mess up events.

models/report.js
mkuklis: @hooverlunch these two functions are only used during the batch mode. I will send you an email with the explanation.

models/report.js
mkuklis: @hooverlunch  lockBatch is an update query and it does not return updated records. I split them into two functions to keep them small but I can move the load batch under lock.

models/report.js
mkuklis: @hooverlunch good idea. I will refactor it.

public/angular/js/services/report.js
mkuklis: @hooverlunch  sounds good!

public/angular/templates/reports/batch.html
mkuklis: @hooverlunch sounds good

hooverlunch:
Great work. A few notes above. Plus:

* The 'xx new reports have arrived' bar should not show up when in batch mode.
* The color of the batch mode intro bar should be the same as the new reports arrived bar. Right now they're slightly different. If possible, please use SASS or LESS or whatever variables to standardize these basic theme colors.
* All text in the table row of an unread report should be boldface font. Both in batch mode and regular mode.
* Checking out a batch should cancel any previous batches for that user (in case they click back and checkout again or something).

public/angular/sass/app.scss
mkuklis: @hooverlunch I think there are two approaches to this. I personally prefer the "un-semantic" approach: keep the classes very generic and combine them to describe different elements. This allows for better class reuse and keeps things simple. I've noticed that Aggie takes different approach so I will refactor.

lib/api/v1/report-controller.js
```diff
@@ -30,6 +31,30 @@ module.exports = function(app, user) {
...
+  // Checkout new batch
+  app.patch('/api/v1/report/batch', user.can('edit reports'), function(req, res) {    
+    batch.checkout(req.session.user._id, function(err, reports) {
+      if (err) res.send(err.status, err.message);
+      else res.send(200, {results: reports, total: reports.length});
+    });
+  });
+
+  // Cancel batch
+  app.put('/api/v1/report/batch', user.can('edit reports'), function(req, res) {
```
hooverlunch: Usually PUT is used to replace a full record. I see you're using it here to distinguish between checkout and cancel but that seems a bit non-standard to me. What about DELETE /report/batch?

lib/api/v1/report-controller.js
```diff
@@ -66,6 +91,44 @@ module.exports = function(app, user) {
     });
   });
 
+  // Mark selected reports as read
+  app.post('/api/v1/report/_read', user.can('edit reports'), function(req, res) {
```
hooverlunch: This should also be PATCH.

lib/api/v1/report-controller.js
```diff
@@ -66,6 +91,44 @@ module.exports = function(app, user) {
...
+        report.save(function(err) {
+          if (err) return res.send(err.status, err.message);
+          if (--remaining === 0) return res.send(200);
+        });
+      });
+    });
+  });
+
+  // Flag selected reports
+  app.post('/api/v1/report/_flag', user.can('edit reports'), function(req, res) {
```
hooverlunch: PATCH here too.

models/batch.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+// helpers
+
+function timeAgo(miliseconds) {
+  var now = new Date();
+  return new Date(now.getTime() - miliseconds);
+}
+
+module.exports = batch;
```
hooverlunch: This looks awesome. Seems like you're treating it as a singleton. If so, can you please make it match the way singletons are done elsewhere in the app? BotMaster is a good example. Just for consistency...

PR #130, opened 2015-01-26
Fixes while load testing

sapraaman:
Performance/debug fixes:

1. Indexes
2. Log exception stack traces
3. Tool enhancement to generate better data for load testing
4. New tool to load test "report" feature
5. Throttling of report queries 
6. Execute find/count in parallel
7. Remove socket handlers when leaving report controller, add them again when report controller is launched


lib/api/streamer.js
```diff
@@ -8,11 +8,12 @@ var _ = require('underscore');
 var util = require('util');
 var EventEmitter = require('events').EventEmitter;
 
-var QUERY_INTERVAL = 100; // 100ms
+var QUERY_INTERVAL = 2000; // 2s
```
hooverlunch: @mkuklis @sapraaman Major reason this is 2s and not 1s? The stats bar is 1s so I thought it would be nice to have them the same. In fact perhaps it could even be a constant in the socket handler.

lib/api/streamer.js
```diff
@@ -8,11 +8,12 @@ var _ = require('underscore');
...
 
 var Streamer = function() {
   this.queries = [];
   this.status = 'idle';
+  this.throttledQuery = _.throttle(this.query, QUERY_INTERVAL);
```
hooverlunch: Why do we need to throttle if we are also doing setInterval?

lib/database.js
```diff
@@ -27,14 +30,14 @@ mongoose.Model.findPage = function(filters, page, options, callback) {
   options.limit = PAGE_LIMIT;
   options.skip = page * PAGE_LIMIT;
 
-  model.count(filters, function(err, count) {
+  async.parallel([
```
hooverlunch: A comment here would be nice. Not clear to me what's going on.

hooverlunch:
Looks good. Just a couple comments.

I am not going to run the tests myself here, so I'm trusting that you guys feel confident that the system can handle the load. Thanks for all your hard work on this!

lib/api/streamer.js
mkuklis: Sounds good @hooverlunch. I will change it to 1s.

lib/api/streamer.js
mkuklis: The throttledQuery was added because currently `query` is being executed every time a new report is being saved.

PR #132, opened 2015-01-27
1787 better add to incident

whitcomb:


public/angular/sass/modals.scss
```diff
@@ -40,3 +48,28 @@
...
+        cursor: pointer;
+        background-color: #eef3f7;
+      }
+    }
+  }
+
+  .modal-footer {
+    margin-top: 0;
+  }
+}
```
hooverlunch: Looks like most of these styles should be common to all modals, not just incident modal. The issue description says "Modal style should be updated and re-used app-wide." Please consider this. Thanks.

hooverlunch:
Looks great! A few notes:

* In future please base pull requests on develop.
* If I click 'Create a new incident' and then 'add to existing', all the links in the modal seem to be broken after that point. Nothing happens when I click them.

public/angular/templates/incidents/index.html
```diff
@@ -56,7 +56,7 @@
     <button class="btn btn-default" aggie-confirm="Are you sure you want to delete selected incidents?" on-confirm="removeSelected()">Delete</a>
   </div>
   <div class="pull-left margin-right padding">
-    <button ng-controller="IncidentFormModalController" class="btn btn-primary" ng-click="create()">Create Incident</a>
+    <button ng-controller="IncidentFormModalController" class="btn btn-info" ng-click="create()">Create new incident</button>
```
hooverlunch: Why did this change? The text should be Title Case, and it now runs off the end of the button. The old wording was better I think.

public/angular/templates/incidents/modal.html
```diff
@@ -22,7 +23,16 @@ <h4 class="modal-title">
...
-          <div class="form-group">
+          <!-- <div class="form-group" ng-hide="minimal">
+            <label for="status" class="col-lg-2 control-label">Status</label>
+            <div class="col-lg-9">
+              <select id="status" name="status" ng-model="incident.status" ng-options="value | capitalize for value in status" class="form-control" required></select>
+              <div class="has-error" ng-show="showErrors">
+                <div class="help-block" ng-show="form.status.$error.required">Status is required.</div>
+              </div>
+            </div>
+          </div> -->
```
hooverlunch: Please remove.

public/angular/templates/incidents/report_incident_modal.html
```diff
@@ -6,24 +6,64 @@ <h4 class="modal-title">
...
+              {{ i.locationName }}
+              <div class="coordinates" ng-show="i.latitude || i.longitude">({{ i.latitude }}, {{ i.longitude }})</div>
+            </td>
+            <td class="compact content">{{ i.assignedTo }}</td>
+            <td class="compact content">{{ i.status | translate }}</td>
+            <td class="compact content">{{ 'veracity.' + i.verified | translate }} </td>
+          </tr>
+        </tbody>
+      </table>
+
```
hooverlunch: The issue description says "Incident table and form code should be re-used here." It looks like you've copied and pasted. I would much prefer actual reuse via an include of some kind. I realize some columns need to be hidden but that shouldn't be terribly difficult. Please refactor. Thanks.

hooverlunch:
The modal is looking a lot better. A few notes:

* In future please base pull requests on develop branch. Please don't actually merge this branch to master, but to develop.
* The incident table does not match the new incident table design. This is likely due to the cut and paste. Should be resolved when refactored as requested above. Still, the as-built table differs significantly from mockup. Not sure why this wasn't noticed.
* If I click 'Create a new incident' and then 'add to existing', all the links in the modal seem to be broken after that point. Nothing happens when I click them.
* The issue description said "Add-to-existing should be shown first, **unless there are no existing incidents**." This is not currently the case.
* I'd like to avoid a full page refresh when an incident is created/selected. Else the user will lose their place in the report list. @mkuklis @sapraaman are we now streaming updates to report attribs? If so then the new incident should appear without needing a page refresh.
* Reports are not getting marked read when report is added to a newly created incident (it works when added to existing) @mkuklis @sapraaman I thought we had this working. If not can someone make this happen please.

There were multiple bits of the issue description that were overlooked here. I try hard to be as detailed and clear as possible in there so please read it carefully.

mkuklis:
@hooverlunch the previous design had only option to choose from existing incidents when assigning them to reports. I haven't looked at the new design but if there is an option to create a new incident on the fly then marking reports as `read`  should be probably added there. @whitcomb here is the old version where this is happening: https://github.com/TID-Lab/aggie/blob/master/public/angular/js/controllers/incidents/report_incident_modal.js#L28

whitcomb:
I believe all of the issues have been addressed with the latest push.

hooverlunch:
This is better. Still a few things are not working right:

* I'm now getting a comma by itself in the location box if there is no lat/lng. Maybe not related to this issue but please fix: 
image
* I'm still getting a page reload when report is added to a **new** incident. Works properly when adding to existing.
* The text 'create a new incident' at the top of the modal should not be title case since it's part of the sentence. You'll note that its counterpart, 'add to an existing incident' remains lowercase. Please revert.
* The add to existing modal is *still* showing first even if there are no incidents. If there are no incidents and someone clicks 'Add', the add-to-new modal should show first.

Please fix these things and then go ahead and merge.

whitcomb:
Merged into develop

PR #137, opened 2015-01-30
1893

mkuklis:
http://redmine.sassafras.coop/issues/1893


mkuklis:
The server side for this issue should be now done.

PR #143, opened 2016-02-04
Graph api

hooverlunch:


lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
   var self = this;
 
   if (!this._url) {
-    process.nextTick(function() { self.emit('error', new Error('Missing Facebook URL')); });
+    process.nextTick(function () {
```
hooverlunch: Our style guide calls for no space between `function` and `()`. See here: https://github.com/felixge/node-style-guide. Thanks.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
-  queries.posts += ' ORDER BY updated_time';
+  if (source.lastIndexOf('/') === source.length - 1) {
+    source = source.substring(0, source.length - 1);
+  }
 
-  // Get new comments associated with posts.
-  queries.comments = "SELECT id, fromid, time, text, post_id FROM comment WHERE (post_id IN (SELECT post_id FROM #posts))";
-  if (this._lastReportDate) queries.comments += ' AND time > ' + (this._lastReportDate.getTime()/1000);
-  queries.comments += ' ORDER BY time';
+  source = source.substring(source.lastIndexOf('/') + 1, source.length);
```
hooverlunch: I would strongly suggest using a regular expression to parse this instead of all these calls to substring, indexOf, etc.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
-    }
-  });
+  if (this._lastReportDate) {
+    params.since = this._lastReportDate.getTime() / 1000;
+  }
+  // Grab posts off of sources feed
+  self._doGet('/v2.2/' + source + '/feed', params, data, callback);
+
+  //If no new posts, grab comments off of most recent post
+  if((!data || data.length == 0) && this.posts[source] != null) {
```
hooverlunch: Space after `if` please.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
+    self._doGet('/v2.2/' + this.posts[source].id + '/comments?order=reverse_chronological', params, data, callback);
```
hooverlunch: I don't quite get how this is supposed to work. Why are you only loading comments if there no new posts? There could be both new posts and new comments on older posts, etc. There are also comment replies, which I think are not being handled here at all... The beauty of FQL was that it allowed grabbing all these things fairly easily I remember. But with the Graph API it might be trickier.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
+    if (object.type) {
+      // Handle each post and any comment already attached to it
+      reportData.push(self._parse(object, {type: 'post'}));
+      if (object.comments && object.comments.data && object.comments.data.length > 0 ) {
+        object.comments.data.forEach(function(comment) {
+          reportData.push(self._parse(comment, {type: 'comment', url: object.link}));
+        });
+      }
+    } else {
+      if (oldPost == undefined){
```
hooverlunch: Some more spacing issues in this file. Please add space after the if condition, and above, please remove space after 0 on line 106.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
+        reportData.push(self._parse(object, {type: 'comment', url: object.link}));
+      } else {
+        reportData.push(self._parse(object, {type: 'comment', url: oldPost.link }));
```
hooverlunch: This line and line 113 are almost entirely duplicates of each other. Please refactor to remove duplication.

test/fixtures/facebook-0.json
```diff
@@ -1,6 +1,3 @@
 {
-  "data": [{
-    "name": "posts",
-    "fql_result_set": []
-  }]
-}
+  "data": []
+}
```
hooverlunch: Please add newline. Also anywhere else in this PR where you see this symbol.

test/fixtures/facebook-1.json
```diff
@@ -1,52 +1,37 @@
...
+            "id": 23232323,
+            "name": "Test User 1"
+          },
+          "message": "Amazing!!!",
+          "created_time": "2015-09-09T20:43:14+0000"
+        }
+      ]
+    }
+  }
+]
```
hooverlunch: Great that you've updated these JSON fixtures to match what the Graph API is returning. I would suggest you double and triple check that this is what is actually being returned as if the fixtures are off then the tests are misleading.

test/fixtures/facebook-1.json
```diff
@@ -1,52 +1,37 @@
...
+  },
+  {
+    "id": 123143123123,
+    "from": {
+      "id": 23232323,
+      "name": "Georgia Tech"
+    },
+    "message": "Bioengineers at Georgia Tech are developing a better way to rebuild and repair broken bones from inside the human body. -- http://c.gatech.edu/1onyDu",
+    "link": "https://www.facebook.com/photo.php?fbid=10154233081380228&set=a.261947725227.302128.251841205227&type=1",
+    "created_time": "2015-09-09T20:43:14+0000",
```
hooverlunch: It looks like you are using the same created_time value for everything. I suggest you change this. When I made these fixtures originally I made a timeline of when the posts and comments were created so that I could be sure everything was getting picked up right. The comments vs. post thing can be confusing. It can be very easy to miss data entirely, and that is a big problem for Aggie.

test/lib.fetching.content-services.facebook-content-service.test.js
```diff
@@ -72,35 +72,29 @@ describe('Facebook content service', function() {
...
   it('should get new comments from old posts', function(done) {
 
-    // Fetch first batch
-    var service = stubWithFixture('facebook-1.json');
-    service.fetch({maxCount: 50}, function(){});
-
-    // Stub for second batch
-    stubWithFixture('facebook-2.json', service);
-
+    var service = stubWithFixture('facebook-2.json');
```
hooverlunch: I'm curious why you removed the "Fetch first batch" block from the old spec. Can you comment?

hooverlunch:
Hi Trey. Thanks so much for your work on this. I think this is pretty close to done. This is a tricky feature and I want to be sure we get this right. I've added a few comments about style but mostly that looks good. I'm more concerned about the logic for fetching comments and making sure we are not missing any data. I left a few questions in there. If you could reply I'd appreciate it.

lib/fetching/content-services/facebook-content-service.js
treywashington: One of the major changes from fql to the graph api I noticed was that comments were moved from being attached to a page to being attached to specific objects (posts) on a page. In order to get all of the new comments I would first have to recursively walk the feed and keep track of every post and then check for comments on each post. To do this we'd have place some depth limit because it gets expensive for multiple sources with long histories of posts. As of right now what I do is only grab new comments from the most recent post of each source. I'm thinking I can easily extend my current implementation to grabbing all new comments from all posts that were gathered from the time fetching was turned on forward.

test/lib.fetching.content-services.facebook-content-service.test.js
treywashington: If I remember correctly I think I found fixtures 1 and 2 redundant so I skipped over the first one in hopes of making a different test later on. Later on is now here, so I'll switch up that fixture and add a few more tests.

treywashington:
It's also worth noting the graph api doesn't allow for large usage rates when compared to twitter api. When we used tested this api on high activity pages we would occasionally reach the user limit and were forced into a 30 minutes "cooldown" period where we couldn't use the api. Facebook recommends we switch out to a different user for the downtime and switchback once the new user limit is reached and repeat. I didn't implement anything to solve this problem.

lib/fetching/content-services/facebook-content-service.js
hooverlunch: I figured this part would be tricky, but here is what I am wondering:

Let's say there is a post p1 created at 2pm and a comment is added at 4pm. If you ask for all posts `since` 3pm, will it include p1 in the returned list -- because it has a comment after 3pm? 

If so, then all you have to do is iterate over _all_ posts _and_ comments in the return set and skip any duplicates. i.e. don't re-add posts you've already added, and don't re-add comments either.

This would be easy to test if you don't already know the answer. I used to set up a fake FB group for the purpose.

test/lib.fetching.content-services.facebook-content-service.test.js
hooverlunch: Right, but I think the reason fixture 1 was loaded first was to test the duplication avoidance stuff I described above.

hooverlunch:
Can you get some hard numbers on the rate limits? We shouldn't need to be hitting it very hard. Just checking for new stuff every few minutes would be fine. If we have some hard numbers we can strategize.

treywashington:
There's app level limiting which allows 200 api calls per user for every 60 mins. "As an example, if your app had 10 users yesterday and 5 new logins today, that would give you a base of 15 users. This means that your app can make ((10 + 5) * 200) = 3000 API calls in any 60 minute window. There's also user level rate limiting which supersedes app limiting and is less specific (this is the one we hit). It says a user accumulates score over time based on calls made. Once the max score, n, is reached calls to that user are blocked for 30 mins and the score decays by n/3 per second.

hooverlunch:
How often were you making requests when you tested?

Let's worry about rate limiting once we have the comment stuff figured out.
I'm curious how many API calls per fetch we're going to end up with.

On Thu, Feb 4, 2016 at 7:16 PM, Trey Washington <notifications@github.com>
wrote:

> There's app level limiting which allows 200 api calls per user for every
> 60 mins. "As an example, if your app had 10 users yesterday and 5 new
> logins today, that would give you a base of 15 users. This means that your
> app can make ((10 + 5) * 200) = 3000 API calls in any 60 minute window.
> There's also user level rate limiting which supersedes app limiting and is
> less specific (this is the one we hit). It says a user accumulates score
> over time based on calls made. Once the max score, n, is reached calls to
> that user are blocked for 30 mins and the score decays by n/3 per second.
>
> —
> <https://github.com/TID-Lab/aggie/pull/143#issuecomment-180118282>.
>



-- 
Tom Smyth

touchstonecohousing.org

lib/fetching/content-services/facebook-content-service.js
treywashington: No this wouldn't fetch p1 or its comment. The graph api only returns small chunks of data at a time so its much more difficult to get all posts, especially with the rate limits.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -26,116 +27,139 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
     case 'comment':
       return {
-        authoredAt: new Date(data.time * 1000),
+        authoredAt: new Date(data.created_time),
         fetchedAt: new Date(),
-        content: data.text,
-        author: this._findAuthor(data.fromid, options.authors),
-        url: options.post.permalink
+        content: data.message,
+        author: data.from.name,
```
andresfib: I am testing the facebook integration with one public page (https://www.facebook.com/CDNH.OFFICIAL/) and the service complains that data has no 'from' key: 
{ message: 'ျပည္သူ႔နီတိ သင္ၾကားေရး၏ ဦးတည္ခ်က္*',
  story: 'CDNH published a note.',
  created_time: '2016-01-28T07:48:09+0000',
  id: '368893233299480_441368276051975' }
Fri, 5 Feb 2016 08:35:04 UTC - debug:  
{ type: 'comment' }
Fri, 5 Feb 2016 08:35:04 UTC - error: TypeError: Cannot read property 'name' of undefined
    at FacebookContentService._parse (/home/amoreno/aggie/lib/fetching/content-services/facebook-content-service.js:153:26)

Where can we find the data formats? Do we need a smarter parser?

treywashington: Are you using a v2.2 api key? I know they switched around the format of the data objects a lot in the new releases

lib/fetching/content-services/facebook-content-service.js
hooverlunch: I just tested this and I think in fact p1 is returned.

Test FB group: https://www.facebook.com/groups/539367472907722/

Graph API request: https://developers.facebook.com/tools/explorer/145634995501895/?method=GET&path=539367472907722%2Ffeed%3Fsince%3D1454685714&version=v2.5

Note how the first post is retrieved while the second one is not. This is because the first post has a comment after the `since` date.

The docs also say the API can return 100 posts at a time. That's not bad in terms of quantity. The tricky part will be fetching the comments. If there were some way to know if a post has new comments, that would save some requests. Perhaps you can look into this.

lib/fetching/content-services/facebook-content-service.js
hooverlunch: I am not sure what you are talking about here.

lib/fetching/content-services/facebook-content-service.js
hooverlunch: Further update: check out this API request:

https://developers.facebook.com/tools/explorer/145634995501895/?method=GET&path=539367472907722%2Ffeed%3Ffields%3Did%2Cmessage%2Cupdated_time%2Ccomments%26since%3D1454689720&version=v2.5

Note that I am asking for the comments field manually and thus getting the comments inline with the request. So no extra requests.

I know this is version 2.5. Is there some reason we are not planning to use 2.5? Support for 2.2 will go away in early 2017.

treywashington: When I started writing the code 2.2 was the latest, but I think it's worth it to go ahead and update to latest while I'm making changes.

lib/fetching/content-services/facebook-content-service.js
andresfib: Thanks, that will explain it for sure. Modifying README.md to reflect the new way to get the new key will help!

andresfib: By the way, when I created a new app (recently) the codes are tied to using API v2.5 and above. Thus, I think we may need an smarter parser after all.

hooverlunch: Note that in v2.5 you have to specify what fields you want. In older versions it would just give you everything.

andresfib: I have now got a 2.0 key and got an instance of aggie graph_api running with 13 different sources. Unsurprisingly the limit was reached quite early. 

More interestingly some of the reports (photos, links,...) have come back with empty content which breaks the angular app because of the Autolinker call

```
var linkify = function(report) {
      report.content = Autolinker.dulink(report.content);
      return report;
    };

```

treywashington:
My newest update handles the empty content error. I'll push what I have for now so you can take a look. I'm thinking the best way around the user limit is to use reduce the rate of fetching on facebook sources or we can use two api keys that rotate out automatically. Any thoughts Tom?

pgarrison:
I've been unable to reproduce the rate limit issue. I've used two different facebook apps, one very old and one brand new, both with zero users. I've tested in Aggie and in a separate little script that just makes tons of facebook api requests. In my own script I've done 1000 requests in ~10 minutes with no problem, as well as 20 requests per minute for many hours. In Aggie I have about 20 facebook sources being checked however frequently Aggie checks.

If this is the only issue remaining on this PR, I'd suggest we merge it unless we can reproduce the rate limit problem reliably.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -9,133 +44,108 @@ var _ = require('underscore');
...
-    reportData.push(self._parse(comment, {type: 'comment', authors: authors, post: postsById[comment.post_id]}));
+  data.forEach(function(post) {
+      // Handle each post and any comments already attached to it
+      reportData.push(self._parse(post));
+      if (post.comments && post.comments.data && post.comments.data.length > 0) {
+        post.comments.data.forEach(function(comment) {
+          comment.link = post.link;
+          reportData.push(self._parse(comment));
+        });
+      }
```
pgarrison: I think generally map + concat should be preferred to each + push, as it's faster and IMO cleaner. Not that performance really matters here though. Also if we have underscore around we might as well use it.

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -1,4 +1,39 @@
...
+  - Service does not use paging because of usage limits.
+
+  - This service will miss any posts beyond 10 that have been made since the last fetch. e.g If 12 posts were made since
+  the last fetch the 1rst and 2nd(oldest) posts will be missed.
+
+  - This service will miss comments on any posts more than 10 posts old. Comments are attached the posts, so no comments
+  on any posts older than the 10th oldest one will be retrieved.
+
+  - It will also miss any comments beyond 10 that have been made since the last fetch.
+*/
```
hooverlunch: This is one of the best comments I have ever read. Well done!

test/lib.fetching.content-services.facebook-content-service.test.js
```diff
@@ -33,6 +33,15 @@ describe('Facebook content service', function() {
...
 
+  it('should grab the correct source given multiple url types', function() {
+    var service = new FacebookContentService({url: "http://test.com/"});
+    expect(service._getSourceFromUrl('http://facebook.com/cocacola')).to.equal('cocacola');
+    expect(service._getSourceFromUrl('https://www.facebook.com/CocaColaUnitedStates/?brand_redir=40796308305&test=2'))
+        .to.equal('CocaColaUnitedStates');
+    expect(service._getSourceFromUrl('https://www.facebook.com/Gatorade/')).to.equal('Gatorade');
+    expect(service._getSourceFromUrl('https://www.facebook.com/groups/Gatorade/')).to.equal('Gatorade');
+  });
+
```
hooverlunch: Nice.

hooverlunch:
Looks great to me. I'm assuming it has been well tested manually and that the specs are passing. In which case, I'd say go ahead and merge!

PR #147, opened 2016-02-26
4000: Media settings can be set up online within aggie application

andresfib:
No need to edit the config file for media settings. Configuration still in file but should be moved to db. Only admin users can change the settings in the app. README updated with new instructions

Interactive test of settings not ready.

### DIFF
--- a/config/secrets.js
+++ b/config/secrets.js
@@ -13,9 +13,7 @@ var fs = require('fs-extra');
 var secretsFile = path.resolve(__dirname, 'secrets.json');
 var data = fs.readFileSync(secretsFile, 'utf8');
 
-// remove comments from secrets file
-var secrets = JSON.parse(jsmin(data));
-nconf.add('secrets', {type: 'literal', store: secrets});
+nconf.add('secrets', {type: 'file', file: secretsFile});
 
 // load server preferences
 var prefsFile = path.resolve(__dirname, 'server-prefs.json');
@@ -56,6 +54,20 @@ module.exports.updateFetching = function(flag, cb) {
   cb = cb || function() {};
   nconf.set('fetching', S(flag).toBoolean());
   nconf.save(function(err){
+    return cb(err);
+  });
+};
+
+// update settings
+module.exports.update = function(type, settings, cb) {
+  cb = cb || function() {};
+
+  for (var key in settings) {
+    var item = type + ':' + key;
+    nconf.set(item, settings[key]);
+  }
+
+  nconf.save(function(err) {
     return cb(err);
   });
 };
--- a/public/angular/js/controllers/media/settings/settings_modal.js
+++ b/public/angular/js/controllers/media/settings/settings_modal.js
@@ -12,8 +12,8 @@ angular.module('Aggie')
 
     $scope.settings = settings;
     $scope.media = media;
-
     $scope._showErrors = false;
+
     $scope.save = function(form, mediaName, settings) {
       settings.configured = true;
       Settings.set(mediaName, settings, success(mediaName, 'saved'), failure);
--- a/public/angular/js/controllers/media/settings/settings_modal.js
+++ b/public/angular/js/controllers/media/settings/settings_modal.js
@@ -13,6 +13,7 @@ angular.module('Aggie')
     $scope.settings = settings;
     $scope.media = media;
     $scope._showErrors = false;
+    $scope.loading = false;
 
     $scope.save = function(form, mediaName, settings) {
       settings.configured = true;
@@ -44,6 +45,31 @@ angular.module('Aggie')
     function failure(data) {
       flash.setAlertNow('An error has occurred');
       console.log('failure: ', data);
+    };
+
+    $scope.test = function(mediaName, settings) {
+      $scope.success = false;
+      $scope.failure = false;
+      $scope.loading = true;
+      Settings.test('media', mediaName, settings, successTest, failureTest);
+    };
+
+    function successTest(response, responseHeaders) {
+      $scope.loading = false;
+
+      if (response.success) {
+        $scope.success = true;
+      } else {
+        $scope.failure = true;
+        $scope.message = response.message;
+      }
+    };
+
+    function failureTest(data) {
+      $scope.loading = false;
+
+      $scope.failure = true;
+      $scope.message = data.message;
     };
   },
 ]);
--- a/public/angular/js/controllers/media/settings/settings_modal.js
+++ b/public/angular/js/controllers/media/settings/settings_modal.js
@@ -1,0 +1,49 @@
+angular.module('Aggie')
+
+.controller('MediaSettingsModalInstanceController', [
+  '$scope',
+  '$modalInstance',
+  'Settings',
+  'FlashService',
+  'settingsValues',
+  'media',
+  '$filter',
+  function($scope, $modalInstance, Settings, flash, settings, media, $filter) {
+
+    $scope.settings = settings;
+    $scope.media = media;
+
+    $scope._showErrors = false;
+    $scope.save = function(form, mediaName, settings) {
+      settings.configured = true;
+      Settings.set(mediaName, settings, success(mediaName, 'saved'), failure);
+      $modalInstance.close();
+    };
+
+    $scope.delete = function(mediaName, settings) {
+
+      for (var setting in settings) {
+        settings[setting] = '';
+      }
+
+      settings.on = false;
+      settings.configured = false;
+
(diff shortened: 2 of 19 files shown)


config/secrets.js
```diff
@@ -59,3 +57,17 @@ module.exports.updateFetching = function(flag, cb) {
...
+
+  for (var key in settings) {
+    var item = type + ':' + key;
+    nconf.set(item, settings[key]);
+  }
+
+  nconf.save(function(err) {
+    return cb(err);
+  });
+};
```
hooverlunch: So I take it that this is temporary code, and will go away once the settings are moved to the DB?

public/angular/js/controllers/media/settings/settings_modal.js
```diff
@@ -0,0 +1,49 @@
...
+  'FlashService',
+  'settingsValues',
+  'media',
+  '$filter',
+  function($scope, $modalInstance, Settings, flash, settings, media, $filter) {
+
+    $scope.settings = settings;
+    $scope.media = media;
+
+    $scope._showErrors = false;
```
hooverlunch: Why is there a blank line above here? Seems the blank should instead be below. Please try to group code into logical 'paragraphs'.

public/angular/js/filters/clean-settings.js
```diff
@@ -0,0 +1,14 @@
...
+
+.filter('cleanSettings', [function() {
+  return function(items) {
+    var filtered = {};
+    angular.forEach(items, function(value, key) {
+      if (key != 'on' && key != 'configured') {
+        filtered[key] = value;
+      }
+    });
+    console.log(filtered);
```
hooverlunch: Please remove.

public/angular/js/filters/clean-settings.js
```diff
@@ -0,0 +1,14 @@
+angular.module('Aggie')
```
hooverlunch: A comment here would be good. It's not clear what this filter is for.

public/angular/templates/media/settings/modal.html
```diff
@@ -0,0 +1,64 @@
...
+      <button class="close" ng-click="close()" aria-hidden="true">×</button>
+      <h4 class="modal-title">
+        <div class="icon-left">&nbsp;Edit {{ media | capitalize }} Settings Source</div>
+      </h4>
+    </div>
+    <div class="modal-body">
+      <alert type="info" ng-show="success">The credentials are valid. You need to save them before
+        continuing.
+      </alert>
+      <alert type="info" ng-show="failure">The credentials are invalid or they have expired</alert>
```
hooverlunch: Please add a period at the end of this sentence for consistency.

public/angular/templates/media/settings/modal.html
```diff
@@ -0,0 +1,64 @@
...
+        </fieldset>
+      </div>
+    </div>
+    <div class="modal-footer">
+      <button class="btn btn-info" aggie-confirm="Are you sure you want to delete the settings for this source?" on-confirm="delete(media, settings)">Delete</button>
+      <button class="btn btn-default" ng-click="close()">Cancel</button>
+      <button class="btn btn-primary" ng-disabled="settingsForm.$invalid" ng-click="save(form, media, settings)">Save &amp; Close</button>
+    </div>
+  </form>
+</div>
```
hooverlunch: This is looking _much_ better!

test/lib.api.v1.settings-controller.test.js
```diff
@@ -4,19 +4,78 @@ var request = require('supertest');
...
+
+  describe('GET /api/v1/settings/twitter', function() {
+    it('should return the settings JSON of twitter', function(done) {
+      request(settingsController)
+        .get('/api/v1/settings/twitter')
+        .expect(200, {twitter: config.get({ reload: true }).twitter, setting: 'twitter' }, done);
+    });
+  });
+
+  describe('POST /api/v1/settings/media/twitter', function() {
```
hooverlunch: I don't think you need to test that certain HTTP methods _don't_ do things. Or maybe I'm missing the point here?

hooverlunch:
This is looking very nice and much better. Bunch of small changes below.

* I don't think we need this message:
image
or for any other switches. It's fine to show one when you change the settings via the modal, but it's a bit overkill for switches I think.
* When I first opened the settings page, the source types were neither off or on:
image
they should always be one or the other.
* There is no reason we should be wrapping these labels:
image
please make the modal wider if need be.
* The test function is nice! However, when it is running, the loading indicator hides most of the modal which looks kind of odd. I suggest an inline loading indicator to the left of the test button.
* The 'invalid or expired' message should not be blue. It should be the failure color.
* This message:
image
is kind of unorthodox and makes for a weird user experience. When you open the settings modal it appears right at the start if you have settings already. I'd say just remove it.
* The delete button shouldn't be blue. Just gray like the others:
image
* The from email field shows weird contents:
image
* Saving the email doesn't work (JS error)
* So there appears to be a bunch of extra non-working stuff in here. Please remove everything but the code for issue 4000.

public/angular/js/controllers/media/settings/settings_modal.js
andresfib: We have different "logics". The blank separates the settings related variables from the other one. It made sense when I was developing, but maybe it doesn't in the big picture of the novel

test/lib.api.v1.settings-controller.test.js
andresfib: I wanted to make sure that in the long run, nobody changes the meaning of the POST method to actually post a new setting, rather than just testing the setting. Maybe better to make a GET request with an extra test toggle?

config/secrets.js
andresfib: Yes, it is temporary.

Toggling fetching also depends on this. We may want to move that setting to the secrets.json and eventually to the database as well.

test/lib.api.v1.settings-controller.test.js
hooverlunch: Oh I didn't realize that's what POST was doing. I think an appropriate RESTful URL would be something like POST settings/media/twitter/test. That way it's a bit clearer what you're doing. I don't think GET would be appropriate.

config/secrets.js
hooverlunch: Yes everything should be in the database. The sooner we can get that done
the better. Currently things are still kind of a mess. So many files in the
config folder.

On Mon, Feb 29, 2016 at 11:05 PM, Andrés Moreno <notifications@github.com>
wrote:

> In config/secrets.js
> <https://github.com/TID-Lab/aggie/pull/147#discussion_r54520462>:
>
> > @@ -59,3 +57,17 @@ module.exports.updateFetching = function(flag, cb) {
> >      return cb(err);
> >    });
> >  };
> > +
> > +// update settings
> > +module.exports.update = function(type, settings, cb) {
> > +  cb = cb || function() {};
> > +
> > +  for (var key in settings) {
> > +    var item = type + ':' + key;
> > +    nconf.set(item, settings[key]);
> > +  }
> > +
> > +  nconf.save(function(err) {
> > +    return cb(err);
> > +  });
> > +};
>
> Yes, it is temporary.
>
> Toggling fetching also depends on this. We may want to move that setting
> to the secrets.json and eventually to the database as well.
>
> —
> <https://github.com/TID-Lab/aggie/pull/147/files#r54520462>.
>



-- 
Tom Smyth

touchstonecohousing.org

hooverlunch:
This is looking great in general. Still some UI issues below to attend to please:

Not sure what happened here with the field names. Please fix them.

image

I like the instructions at the top of the modal. They are a bit dense though. Could you just quickly increase the `line-height` to give them a bit of breathing room?

image

Sorry I just noticed this one:

image

Please just remove the test button in this case. It is a bad UX to offer a button that then says it doesn't work.

Why does ELMO have authToken and accessToken? It should only have one or the other.

image

Remind me why there is a 'Save' button for fetching? Weren't we going to get rid of that?

image

I have not tested whether the settings that are getting saved here are actually getting used by the fetching module. I am trusting you all on this one. Please double check to make sure everything is hooked up properly.

I think this is good to merge once the above are fixed. Thanks.

andresfib:
Thanks for your comments.

I have fixed them now. 3 points:
1. I am in the process of implementing emitter and listeners to make sure bots are recreated when the settings change.
2. The settings options are collected from the secrets.json. Could you check that yours does not have any extra one?
3. The Save button has already disappeared. Branch already merged in develop, [and it is the reason of the conflict]. It will not come back!

hooverlunch:
Cool, thx.

PR #148, opened 2016-02-29
3939: Admin now has to change pwd in modal after first login

andresfib:
Admin must change his password in a modal that opens from navbar.js.
New field has been added to user model to record whether the user has the default password.


pgarrison:
Do we want to add tests for this?

hooverlunch:
The code is looking good, and I agree with Philip's comments.

As for the UX, I actually think this doesn't make sense as a modal, but as a separate page you get shown after you log in with the temp password. Can you refactor?

Also, can you please standardize the place where form status messages are shown. In 4000 you have it at the top of the form, and in this one it's at the bottom.

Thanks!

public/angular/templates/admin_choose_password_modal.html
```diff
@@ -0,0 +1,34 @@
+<div>
+  <div class="modal-header">
```
hooverlunch: Why is this file called '..._modal' and still has modal styles? Refactor?

hooverlunch:
Looks much better as page. Thanks for that change.

See comments above. Some other issues:
* Submitting form with both boxes empty raises a bunch of JS errors, and actually appears to change the password since I can't login with 'letmein' after doing that. So one accidental click on the submit button makes the app unusable. Please give a little more attention to testing and debugging before submitting changes like this.
* Menu links except for perhaps logout should not be shown. Right now I can click some of them and others do nothing (return 403 in the background). This is not a good user experience.
* The way the validation errors are presented is unorthodox. Typically forms only show validation errors if the user has entered something. Here they show up right at the beginning.
* Can we please up the min characters to 8. 6 is short nowadays.
* After I click 'confirm' with correct passwords, the loading indicator shows and I am returned to the same page again. So this is totally broken. Is this working for you?

hooverlunch:
This looks much better now, thanks.

But I am still getting an error. Here are my steps:

1. Delete the database
1. `node install.js` (creates new user)
1. Login with new admin user
1. Enter password and click confirm -> the loading indicator shows for a second, and then the form comes up again, still filled out (?)
1. Click confirm again -> "There was an error setting the new admin password."

Please try these exact steps. If it doesn't happen to you, maybe it's something with my config.

pgarrison:
I followed your steps and didn't have a problem. After making the new password, I am redirected to the reports page. Check that you are using the right config file (copied from `config/secrets.json.example`)?

andresfib:
Tom, do you have any thoughts on asking the admin to set its email address in this same screen?

hooverlunch:
Ok, that's fine, I trust you all. Just checking.

On Thu, Apr 21, 2016 at 5:13 AM, Philip Garrison <notifications@github.com>
wrote:

> I followed your steps and didn't have a problem. After making the new
> password, I am redirected to the reports page. Check that you are using the
> right config file (copied from config/secrets.json.example)?
>
> —
> You are receiving this because you commented.
> <https://github.com/TID-Lab/aggie/pull/148#issuecomment-212823575>
>



-- 
Tom Smyth

touchstonecohousing.org

hooverlunch:
With what purpose? I think we should keep this simple for now. We can add
that later if it's needed.

On Thu, Apr 21, 2016 at 5:15 AM, Andrés Moreno <notifications@github.com>
wrote:

> Tom, do you have any thoughts on asking the admin to set its email address
> in this same screen?
>
> —
> You are receiving this because you commented.
> <https://github.com/TID-Lab/aggie/pull/148#issuecomment-212824479>
>



-- 
Tom Smyth

touchstonecohousing.org

PR #149, opened 2016-02-29
4137: Removed button to save settings and refactored fetching controller

andresfib:
Switch to set the fetching state changes the state automatically without need to push the Save button, which has been removed


hooverlunch:
This looks fine. Thanks.

PR #150, opened 2016-03-02
4155: Settings for source types can now be tested online

andresfib:
The settings modal has one Test button for the purpose


pgarrison:
I think we should also change, on line 28, `res.send(500);` to `return res.send(500);`

PR #151, opened 2016-03-09
4208 allow numbers in nicknames

pgarrison:
In the legacy version of `mongoose-validator`, `validate('max', 20)` would fail on strings that begin with a number greater than 20. Updating `mongoose-validator` solves this.


pgarrison:
Doing this on develop instead..

PR #153, opened 2016-03-11
4187 https support

andresfib:
HTTPS now is the only way to access Aggie. Important when API credential information is transmitted between client and server.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -5,6 +5,7 @@ process.title = 'aggie-api';
 
 var childProcess = require('./child-process');
 var path = require('path');
+var fs = require('fs');
 var https = require('https');
 var logger = require('./logger');
 var config = require('../config/secrets').get();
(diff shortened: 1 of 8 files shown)


pgarrison:
The code looks good to me. I'd like to hear feedback from @hooverlunch just to make sure this is what we want to be doing.

lib/api.js
```diff
@@ -16,10 +17,17 @@ var exec = require('child_process').exec;
 // Extend global error class
 require('./error');
 
+// Get full path for certificate files
+var keyFile = path.resolve(__dirname, '../config/key.pem');
+var certFile = path.resolve(__dirname, '../config/cert.pem');
+
```
hooverlunch: Will the app no longer work without https?

hooverlunch:
Looks great to me. Thanks.

PR #154, opened 2016-03-21
4154 email settings online

andresfib:


hooverlunch:
@andresfib @pgarrison What is the status of this PR? It doesn't have a label...

pgarrison:
Appropriately unlabeled, I think. It depends on some other work in progress Andres is doing on the settings page.

andresfib:
This is now should be merged into develop as branch 4000 has been merged there already.  
Branch 4247 emits an event when settings are updated. I will first rebase this branch (4154) with that branch (4247) and improve the implementation of email settings reloading.

PR #155, opened 2016-04-13
Fix up process manager tests

pgarrison:
Due to a storm of side effects from other tests, the final test in this
suite was passing `npm test` when it should have been failing. Two of
the lines in this change implement the fix to make this test suite run
successfully in isolation--the others are just for clarity.

1. `+require('./init');`
   Gotta connect to the database.

2. Replace `type` with `media`
   Type was renamed to media as of 35fd74


hooverlunch:
Nice analysis and comment. Thanks.

PR #157, opened 2016-04-13
ESLint configuration

pgarrison:
Adds config files for ESLint, with the idea that this might eventually be enforced by CI tools. The content of these rules is totally up for debate, but I am of the opinion that having well-defined style rules is better than not.

The process for picking the rules was a combination of following the [node-style-guide](https://github.com/felixge/node-style-guide), quantitative comparisons of which styles are used in current code, and personal preference.

These rules are almost a superset of the node-style-guide, except for the rule
```"object-curly-spacing": ["error", "always"],```
In this case, node-style-guide has the opposite, "never". The version of the rule included here is more common in our codebase, and looks better in my opinion.

Most problematically, in my opinion, these rules are not followed by our current codebase. ESLint reports 1978 errors and 291 warnings. On the order of 700 of these can be automatically fixed (eslint --fix is a fabulous tool), but the remainder would need manual attention if we want to get the whole codebase in shape.

These rules are for eslint version 2 (currently 2.4.0).

If we ever advance the project to a version of node that supports ES2015, we want to add new rules for the new syntax. In addition, the `no-inner-declarations` rule will be out of date.


hooverlunch:
Very cool, thanks!

PR #158, opened 2016-04-14
4247 propagate settings change

andresfib:
Changes to the settings of media (facebook, twitter,..) via the app will reload the configuration options in the bots, removing the need to restart the server.


config/secrets.json.example
```diff
@@ -3,19 +3,13 @@
     "consumer_key": "Enter Consumer Key",
     "consumer_secret": "Enter Consumer Secret",
     "access_token": "Enter Access Token",
-    "access_token_secret": "Enter Access Token Secret",
-    "configured": false,
-    "on": false
+    "access_token_secret": "Enter Access Token Secret"
```
hooverlunch: Why are these still here? I thought this is done online now?

lib/fetching/content-services/elmo-content-service.js
```diff
@@ -2,15 +2,17 @@
...
 var util = require('util');
-var config = require('../../../config/secrets').get().elmo;
+var config = require('../../../config/secrets');
 var request = require('request');
 var dateFormat = require('dateformat');
 var _ = require('underscore');
 
 // options.url - The URL of the ELMO api endpoint.
 // options.lastReportDate - The fetchedAt time of the last already fetched report (optional).
+// options.authToken - This value is only passed when testing
```
hooverlunch: Nice comment.

lib/fetching/content-services/dummy-content-service.js
```diff
@@ -52,4 +54,8 @@ DummyContentService.prototype._emitNext = function() {
     setTimeout(function() { self._emitNext(); }, 1);
 };
 
+DummyContentService.prototype._reloadSettings = function() {
```
hooverlunch: This is not a private method since it's called from outside. So it shouldn't start with _

lib/fetching/content-services/elmo-content-service.js
```diff
@@ -2,15 +2,17 @@
...
 var ELMOContentService = function(options) {
-  this._authToken = options.authToken || (config && config.authToken);
+  this._authToken = options.authToken ||
+    (config.get().elmo && config.get({ reload: true }).elmo.authToken);
```
hooverlunch: Why are you reloading here?

lib/fetching/bot-master.js
```diff
@@ -176,6 +189,9 @@ BotMaster.prototype._add = function(bot) {
   var self = this;
   this.bots.push(bot);
   this.emit('botMaster:addedBot');
+  this.on('settingsUpdated:' + bot.source.media, function() {
+    bot.contentService._reloadSettings();
```
hooverlunch: I have a suggestion: emit an event called just, `settingsUpdated` and pass a param indicating the media type. Then, catch that event in one place per process and reload the settings object once. Then let the bots also watch for that event and read from the updated settings object if appropriate.

Currently, if you have 10 Facebook bots and someone updates the facebook auth token, your code will (I think) reload the settings file 10 times...

models/source.js
```diff
@@ -22,8 +22,10 @@ var url_validator = validate({
   passIfEmpty: true
 });
 
+var mediaValues = ['facebook', 'elmo', 'twitter', 'rss', 'dummy'];
+
 var sourceSchema = new mongoose.Schema({
-  media: String,
+  media: {type: String, enum: mediaValues},
```
hooverlunch: This is neat.

public/angular/templates/media/settings/modal.html
```diff
@@ -46,7 +46,7 @@ <h4 class="modal-title">
       <p ng-show="settingsForm.$invalid">The field is required.</p>
       <div class="well bs-component">
         <fieldset>
-          <div class="form-group" ng-repeat="(settingName, settingValue) in settings | cleanSettings" >
+          <div class="form-group" ng-repeat="(settingName, settingValue) in settings" >
```
hooverlunch: Extra space at end of tag.

public/angular/templates/settings.html
```diff
@@ -4,65 +4,57 @@
...
-                   class="toggle-item ng-class:{'active':on===true}">
-              On
-            </label>
-          </div>
-        </td>
-      </tr>
-    </tbody>
-  </table>
+<div class="row" ng-controller="FetchingController">
+  <div class="col-sm-5">
```
hooverlunch: What is this view code doing in this PR? Was this left over from another feature, or?

hooverlunch:
This is looking nice. Thank you. See my suggestion about event naming and a few other small things.

andresfib:
Thanks Tom, I will fix this before moving to branch 4154 email settings online.

Regarding the event passing, you are right. If there are ten bots, Aggie will reload the config file 10 times. So yes, it will be better if the file reloading only happens one per process. Rather than listening for the same event, I should emit another event within that process after the reload has happened, so that the bots get the updated config, right?

I have had problems emitting events with parameters in aggie. The event would die silently and no reach to listeners in other processes. Insights are welcome!

hooverlunch:
I'm not sure about the events with no params. That's unfortunate. Try to
debug it?

On Thu, Apr 21, 2016 at 4:47 AM, Andrés Moreno <notifications@github.com>
wrote:

> Thanks Tom, I will fix this before moving to branch 4154 email settings
> online.
>
> Regarding the event passing, you are right. If there are ten bots, Aggie
> will reload the config file 10 times. So yes, it will be better if the file
> reloading only happens one per process. Rather than listening for the same
> event, I should emit another event within that process after the reload has
> happened, so that the bots get the updated config, right?
>
> I have had problems emitting events with parameters in aggie. The event
> would die silently and no reach to listeners in other processes. Insights
> are welcome!
>
> —
> You are receiving this because you commented.
> <https://github.com/TID-Lab/aggie/pull/158#issuecomment-212816114>
>



-- 
Tom Smyth

touchstonecohousing.org

config/secrets.json.example
andresfib: The settings can still be configured in the secrets.json. And we still use that file as storage for the settings, even if modified online

public/angular/templates/settings.html
andresfib: There has been a small style modification

PR #159, opened 2016-04-15
4154 email settings online

andresfib:
This depends on 4247 branch. First 5 commits belong to that branch.


README.md
```diff
@@ -109,7 +109,7 @@ You can adjust the settings in the `config/secrets.json` file to configure the a
 ### Emails
   1. `fromEmail` is the email address from which system emails come. Also used for the default admin user.
   1. `email.from` is the address from which application emails will come
-  1. `email.transport` is the set of parameters that will be passed to [NodeMailer](http://www.nodemailer.com).
+  1. `email.transport` is the set of parameters that will be passed to [NodeMailer 2.1.0](http://www.nodemailer.com). Valid transport method values are: 'SES', 'sendgrid' and 'SMTP'.
```
hooverlunch: These shouldn't still be in the readme...

config/secrets.js
```diff
@@ -45,7 +43,12 @@ fs.ensureFileSync(_configuration.logger.analytics.filename);
 
 // return configuration
 module.exports.get = function(options) {
-  if (options && options.reload) _configuration = nconf.get();
+  if (options && options.reload) {
+
```
hooverlunch: No blank line here please.

lib/mailer.js
```diff
@@ -1,27 +1,61 @@
...
+var config = require('../config/secrets');
 var _ = require('underscore');
+var logger = require('./logger');
+var ses = require('nodemailer-ses-transport');
+var sendGrid = require('nodemailer-sendgrid-transport');
+
+function Mailer() {
+  this.config = config.get();
+  this.reloadTransport();
+
```
hooverlunch: No blank line please.

lib/mailer.js
```diff
@@ -1,27 +1,61 @@
...
+    method = ses;
+    break;
+  case 'SendGrid':
+    method = sendGrid;
+    options.auth = { api_key: options.api_key };
+    break;
+  case 'SMTP':
+    // Creating transport for SMTP does not require special method, just the options
+    options.auth = { user: options.user, pass: options.pass };
+    method = _.identity;
```
hooverlunch: This is a nice touch :)

public/angular/js/controllers/email.js
```diff
@@ -0,0 +1,51 @@
...
+      flash.setNoticeNow('The mail setting has been correctly saved');
+    }
+
+    $scope.saveEmail = function() {
+      Settings.set('email', { from: $scope.email, transport: $scope.transport }, success, failure);
+    };
+
+    $scope.editTransport = function() {
+      Settings.set('email', { from: $scope.email }, success, failure);
+    };
```
hooverlunch: Is this function supposed to be here?

public/angular/js/controllers/email_modal.js
```diff
@@ -0,0 +1,50 @@
...
+  function($scope, $modalInstance, Settings, emailTransportOptions, flash, settings) {
+    $scope.transport = {};
+    $scope.transport.method = settings.method || {};
+    $scope.transport.options = settings.options || {};
+    $scope.transport.from = settings.from || {};
+    $scope._showErrors = false;
+    $scope.emailTransportOptions = emailTransportOptions;
+
+    $scope.save = function() {
+
```
hooverlunch: No blank please.

public/angular/js/controllers/email_modal.js
```diff
@@ -0,0 +1,50 @@
...
+      $modalInstance.close($scope.transport.method);
+    };
+
+    $scope.close = function() {
+      $modalInstance.dismiss('cancel');
+    };
+
+    function setSetting() {
+
```
hooverlunch: No blank please/

public/angular/templates/email_modal.html
```diff
@@ -0,0 +1,45 @@
...
+                  <option value="">Choose transport method</option>
+                </select>
+                <div class="has-error" ng-show="showErrors()">
+                  <div class="help-block" ng-show="form.method.$error.required">Select transport first</div>
+                </div>
+              </div>
+            </div>
+
+            <div ng-repeat="setting in emailTransportOptions[transport.method]">
+              <label for="{{ setting }}" class="col-lg-2 control-label">{{ setting }}</label>
```
hooverlunch: I'd like to have more human friendly labels for these fields. The best way to do this would be to use the i18n stuff, but it's not ready, so I'll put an issue in the backlog for now.

hooverlunch:
There should be spaces between these fields, as with the edit media settings:

image

Else it looks like one giant box.

public/angular/js/controllers/email.js
```diff
@@ -0,0 +1,51 @@
...
+    Settings.get('email', function success(data) {
+      $scope.email = data.email.from;
+      $scope.transport = angular.copy(data.email.transport);
+    }, failure);
+
+    function success() {
+      flash.setNoticeNow('The mail setting has been correctly saved');
+    }
+
+    $scope.saveEmail = function() {
```
hooverlunch: I fear this is not a very scalable approach to handling settings. You will have one method for each saveable setting. Perhaps fine for now but you will want to refactor once more are added.

hooverlunch:
Please remove all table headers from the page like this:

image

I don't think they are necessary.

hooverlunch:
I just checked the [nodemailer docs](https://nodemailer.com/2-0-0-beta/setup-smtp/) and it looks like the SMTP settings are supposed to look like this for SMTP:

```
var smtpConfig = {
    host: 'smtp.gmail.com',
    port: 465,
    secure: true, // use SSL
    auth: {
        user: 'user@gmail.com',
        pass: 'pass'
    }
};
```

The app gives me this:

```
{
      "method": "SMTP",
      "options": {
        "host": "smtp.webfaction.com",
        "port": 587,
        "secure": "true",
        "user": "hi",
        "pass": "bob"
      }
```

Note the differences: true is quoted, meaning it will be impossible to set this to false (it will be "false"); and user and pass are not inside auth.

This means it probably won't work, and this means that this code wasn't even tested.

I have said it before, and I'll say it again now with emphasis *it is imperative that all features be tested before they are submitted for review*.

There is no point in me even reviewing them otherwise. If I hadn't caught this, there would be broken code in the codebase. The next time some user went to setup SMTP, it wouldn't work. This is not acceptable.

So I'm asking that you please test *all three* of these transport methods and resubmit.

hooverlunch:
This should be a checkbox mapping to true/false, not a text box:

image

hooverlunch:
Please make the save button appear only if the value is changed:

image

I think there is an easy way to do this in Angular...

hooverlunch:
I have refactored some of the settings view code. Please take a look at my changes and let me know if you have comments or questions.

README.md
andresfib: People can still modify the secrets.json file if they want

PR #160, opened 2016-04-18
Shrinkwrap dependencies

pgarrison:
Lock down dependency versions for repeatable testing.


hooverlunch:
Hooray!

pgarrison:
I should mention, for reference, that this change caused me to have to update npm from the latest stable release on debian (1.4.21), because of npm/npm/issues/5920. For reference, I'm now using npm version 3.8.6 without issues.

PR #162, opened 2016-04-18
4396: Http server is created if not certs

andresfib:
HTTPS server is not created if no certs are found. Instead normal HTTP
is created.

Admin is informed in the settings page.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -6,7 +6,6 @@ process.title = 'aggie-api';
 var childProcess = require('./child-process');
 var path = require('path');
 var fs = require('fs');
-var https = require('https');
 var logger = require('./logger');
 var config = require('../config/secrets').get();
 var S = require('string');
@@ -24,10 +23,21 @@ var certFile = path.resolve(__dirname, '../config/cert.pem');
 // Start express server
 var express = require('express');
 var app = express();
-var server = https.createServer({
-  key: fs.readFileSync(keyFile),
-  cert: fs.readFileSync(certFile),
-}, app);
+var server;
+
+// We default to https protocol, if certs are not found fallback to http
+var  protocol = 'https';
+try {
+  server = require('https').createServer({
+    key: fs.readFileSync(keyFile),
+    cert: fs.readFileSync(certFile),
+  }, app);
+} catch (e) {
+  server = require('http').createServer(app);
+  protocol = 'http';
+
+}
+
 
 // Handle request time outs, return 500 in case of timeouts
 function handleRequestTimeouts(req, res, next) {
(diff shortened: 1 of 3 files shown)


lib/api.js
```diff
@@ -24,10 +23,21 @@ var certFile = path.resolve(__dirname, '../config/cert.pem');
...
+try {
+  server = require('https').createServer({
+    key: fs.readFileSync(keyFile),
+    cert: fs.readFileSync(certFile),
+  }, app);
+} catch (e) {
+  server = require('http').createServer(app);
+  protocol = 'http';
+
+}
```
hooverlunch: I don't think try/catch is the best approach here b/c you could end up failing over to http due to some unrelated error. Instead I would test for existence of just the `certFile`. If it's there, do https, if not, do http. If the key is missing or cert is invalid or whatever, it will fail loudly, which is appropriate.

lib/api.js
```diff
@@ -222,6 +232,7 @@ app.set('port', process.env.PORT || 3000);
 
 server.listen(app.get('port'), function() {
   logger.info("✔ Aggie is listening on port " + app.get('port'));
+  logger.info("✔ Aggie is listening to protocol " + protocol);
```
hooverlunch: Is there a reason to remove the port number? Might be useful for those on http. Perhaps just add the protocol in addition?

public/angular/js/controllers/media/settings.js
```diff
@@ -19,6 +20,10 @@ angular.module('Aggie')
       var modalInstance = modal.create(media, $scope.media[media]);
     };
 
+    $scope.inHttp = false;
+    if ($window.location.protocol === 'http:') {
+      $scope.inHttp = true;
+    }
```
hooverlunch: You can simplify these 4 lines to `$scope.inHttp = ($window.location.protocol === 'http:')`

public/angular/templates/settings.html
```diff
@@ -28,6 +28,12 @@
 </div>
 
 <div ng-controller="MediaSettingsController">
+  <div class="alert alert-danger" role="alert" ng-if="inHttp">
+    <span>Media settings modified here will be unsecuredly transmitted to the server.
+      We strongly advise to <a href="https://github.com/TID-Lab/aggie#installation">
+      <span class="text-blue">install</span></a> the SSL
+      certificates in the server.</span>
+  </div>
```
hooverlunch: Nice touch!

lib/api.js
pgarrison: Looks like that's what it's doing already.

lib/api.js
pgarrison: Still going to use try/catch, but only checking for ENOENT from the `certFile`. The alternative is to use the asynchronous version, but a decent chunk of the code below would have to go in the callback.

pgarrison:
Before I merge this, I want to note that we don't have tests cases for this code (of course I've checked it manually, but it might break quietly in the future).

PR #166, opened 2016-05-19
4142 internationalization

pgarrison:
Nearly all of the Aggie interface can now be translated into other languages. Translation files are in `public/angular/translations/locale-*.json` (and for emails in `lib/mailer.js`). Info on how to add new translations is in `contributing.md`. 

Some things that should be translated but can't be (yet)
* The strings `* Any Incident` and `* Without Incident` in the reports page for filtering reports by incident.
* Errors messages that originate on the server
  - These can easily be translated in the same manner as the rest of the strings: just replace the string with a translation id and then add the `"id": "string"` pair to the translation files.
* Tooltips (mainly in trends)


pgarrison:
@andresfib can you translate the emails to spanish for real (just so we have a second language there as an example)?

lib/mailer.js
```diff
@@ -59,29 +60,55 @@ Mailer.prototype.send = function(subject, body, to, from, callback) {
...
-    body: '<p>To change your Aggie password, <a href="<%= protocol %>://<%= host %>/password_reset/<%= token %>">click here</a> or copy the link below into your browser.</p><p><%= protocol %>://<%= host %>/password_reset/<%= token %></p><p>You are receiving this message because someone requested a new password from <%= host %> for an account registered with this email address. If you did not initiate this request, please ignore this email.</p>'
+  es: {
+    newUser: {
+      subject: 'Bienvenido a Aggie!',
+      body: 'Algo en Español'
+    },
+    forgotPassword: {
+      subject: 'Aggie Contraseña Reiniciar Solicitud',
+      body: '<p>En Español</p>'
+    }
```
hooverlunch: Should be in separate file please.

public/angular/translations/locale-debug.json
```diff
@@ -0,0 +1,353 @@
...
+  "Read": "READ",
+  "Unread": "UNREAD",
+  "Flagged": "FLAGGED",
+  "Unflagged": "UNFLAGGED",
+  "Read & Unflagged": "READ & UNFLAGGED",
+  "Total Reports": "TOTAL REPORTS",
+  "Reports/min": "REPORTS/MIN",
+  "Escalated Incidents": "ESCALATED INCIDENTS",
+  "Mark all Unread": "MARK ALL UNREAD"
+}
```
hooverlunch: Perhaps add this to .gitignore and put steps to generate (gulp script) and use in Readme file.

test/public.angular.translations.test.js
```diff
@@ -0,0 +1,95 @@
...
+      if (err) return callback(err);
+      callback(null, _.object(allLanguages));
+    });
+  });
+}
+
+describe('Translations files', function() {
+  it('debug language should contain the union of the other languages',
+     function(done) {
+    getTranslations(function(err, allLanguages) {
```
hooverlunch: Indent issues here.

pgarrison: Not a problem as far as I can tell (this diff is outdated but the same thing appears in the new diff), just a little odd-looking.

pgarrison:
Translations for emails still have no tests.

pgarrison:
After poking around a little with debug translations on, I would estimate that around 80% of the strings are translated with this. Hopefully we can gradually the rest of the missing ones as we find them.

PR #167, opened 2016-06-17
1785 api authorization

andresfib:
Implemented authorization as per 
https://docs.google.com/spreadsheets/d/1Rv_SMIKfNhv4yIDWsJ6ibESKlRcQXa1rUieE1hSu1hk/edit#gid=0

In the process:
Simplified permissions
Tweaked UI for viewers
Special arrangements done to allow for viewers and monitors to modify their user information

### DIFF
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -48,10 +48,10 @@ module.exports = function(app, user) {
 
   // Get a User by id
   app.get('/api/v1/user/:_id', user.can('admin users'), function(req, res) {
-    User.findById(req.params._id, '-password', function(err, u) {
+    User.findById(req.params._id, '-password', function(err, user) {
       if (err) res.send(err.status, err.message);
-      else if (!u) res.send(404);
-      else res.send(200, u);
+      else if (!user) res.send(404);
+      else res.send(200, user);
     });
   });
 
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -2,14 +2,15 @@
 
 var express = require('express');
 var User = require('../../../models/user');
+var UserPermissions = require('../../../shared/user.js');
 var mailer = require('../../mailer');
 var _ = require('underscore');
 
 module.exports = function(app, user) {
   app = app || express();
-  user = user || require('../authorization')(app);
   var auth = require('../authentication')(app);
   var password = require('../reset-password')(app, auth);
+  user = user || require('../authorization')(app);
 
   function sendEmail(user, req, callback) {
     // Only send email if not in test mode.
@@ -37,7 +38,7 @@ module.exports = function(app, user) {
   // Get a list of all Users
   app.get('/api/v1/user', user.can('view users'), function(req, res) {
     var query = {};
-    if (req.user && req.user.role === 'viewer')
+    if (req.user && (new UserPermissions(req.user)).can('not view other users'))
       query.username= req.user.username;
 
     User.find(query, '-password', function(err, users) {
@@ -76,7 +77,7 @@ module.exports = function(app, user) {
       if (err) return res.send(err.status, err.message);
       if (!user) return res.send(404);
       //Only admin can update users other than itself
-      if (req.user && req.user.role !== 'admin' && req.params._id != req.user._id) return res.send(403);
+      if (req.user && !(new UserPermissions(req.user)).can('admin users') && req.params._id != req.user._id) return res.send(403);
 
       for (var attr in req.body) {
         user[attr] = req.body[attr];
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -37,7 +37,7 @@ module.exports = function(app, user) {
   // Get a list of all Users
   app.get('/api/v1/user', user.can('view users'), function(req, res) {
     var query = {};
-    if (req.user.role === 'viewer')
+    if (req.user && req.user.role === 'viewer')
       query.username= req.user.username;
 
     User.find(query, '-password', function(err, users) {
@@ -76,7 +76,7 @@ module.exports = function(app, user) {
       if (err) return res.send(err.status, err.message);
       if (!user) return res.send(404);
       //Only admin can update users other than itself
-      if (req.user.role !== 'admin' && req.params._id != req.user._id) return res.send(403);
+      if (req.user && req.user.role !== 'admin' && req.params._id != req.user._id) return res.send(403);
 
       for (var attr in req.body) {
         user[attr] = req.body[attr];
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -35,7 +35,7 @@ module.exports = function(app, user) {
   app.use(express.json());
 
   // Get a list of all Users
-  app.get('/api/v1/user', user.can('manage users'), function(req, res) {
+  app.get('/api/v1/user', user.can('view users'), function(req, res) {
     User.find({}, '-password', function(err, users) {
       if (err) res.send(err.status, err.message);
       else res.send(200, users);
@@ -43,7 +43,7 @@ module.exports = function(app, user) {
   });
 
   // Get a User by id
-  app.get('/api/v1/user/:_id', user.can('manage users'), function(req, res) {
+  app.get('/api/v1/user/:_id', user.can('view users'), function(req, res) {
     User.findById(req.params._id, '-password', function(err, user) {
       if (err) res.send(err.status, err.message);
       else if (!user) res.send(404);
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -36,23 +36,27 @@ module.exports = function(app, user) {
 
   // Get a list of all Users
   app.get('/api/v1/user', user.can('view users'), function(req, res) {
-    User.find({}, '-password', function(err, users) {
+    var query = {};
+    if (req.user.role === 'viewer')
+      query.username= req.user.username;
+
+    User.find(query, '-password', function(err, users) {
       if (err) res.send(err.status, err.message);
       else res.send(200, users);
     });
   });
 
   // Get a User by id
-  app.get('/api/v1/user/:_id', user.can('view users'), function(req, res) {
-    User.findById(req.params._id, '-password', function(err, user) {
+  app.get('/api/v1/user/:_id', user.can('admin users'), function(req, res) {
+    User.findById(req.params._id, '-password', function(err, u) {
       if (err) res.send(err.status, err.message);
-      else if (!user) res.send(404);
-      else res.send(200, user);
+      else if (!u) res.send(404);
+      else res.send(200, u);
     });
   });
 
   // Create a new User
-  app.post('/api/v1/user', user.can('manage users'), function(req, res) {
+  app.post('/api/v1/user', user.can('admin users'), function(req, res) {
     User.create(req.body, function(err, user) {
       err = Error.decode(err);
       if (err) res.send(err.status, err.message);
@@ -67,10 +71,12 @@ module.exports = function(app, user) {
   });
 
   // Update a User
-  app.put('/api/v1/user/:_id', user.can('manage users'), function(req, res) {
+  app.put('/api/v1/user/:_id', user.can('update users'), function(req, res) {
     User.findById(req.params._id, function(err, user) {
       if (err) return res.send(err.status, err.message);
       if (!user) return res.send(404);
+      //Only admin can update users other than itself
+      if (req.user.role !== 'admin' && req.params._id != req.user._id) return res.send(403);
 
       for (var attr in req.body) {
         user[attr] = req.body[attr];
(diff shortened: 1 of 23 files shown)


lib/api/v1/user-controller.js
```diff
@@ -34,24 +34,28 @@ module.exports = function(app, user) {
...
   // Get a User by id
-  app.get('/api/v1/user/:_id', user.can('manage users'), function(req, res) {
-    User.findById(req.params._id, '-password', function(err, user) {
+  app.get('/api/v1/user/:_id', user.can('admin users'), function(req, res) {
+    User.findById(req.params._id, '-password', function(err, u) {
       if (err) res.send(err.status, err.message);
-      else if (!user) res.send(404);
-      else res.send(200, user);
+      else if (!u) res.send(404);
+      else res.send(200, u);
```
pgarrison: I think `user` was better

lib/api/v1/user-controller.js
```diff
@@ -34,24 +34,28 @@ module.exports = function(app, user) {
   app.use(express.json());
 
   // Get a list of all Users
-  app.get('/api/v1/user', user.can('manage users'), function(req, res) {
-    User.find({}, '-password', function(err, users) {
+  app.get('/api/v1/user', user.can('view users'), function(req, res) {
+    var query = {};
+    if (req.user && req.user.role === 'viewer')
+      query.username= req.user.username;
```
pgarrison: This is really a bit of a hack. Instead of checking that the role is `viewer`, we should really be checking that the user does not have the `admin users` permission.

lib/api/v1/user-controller.js
```diff
@@ -66,10 +70,12 @@ module.exports = function(app, user) {
   });
 
   // Update a User
-  app.put('/api/v1/user/:_id', user.can('manage users'), function(req, res) {
+  app.put('/api/v1/user/:_id', user.can('update users'), function(req, res) {
     User.findById(req.params._id, function(err, user) {
       if (err) return res.send(err.status, err.message);
       if (!user) return res.send(404);
+      //Only admin can update users other than itself
+      if (req.user && req.user.role !== 'admin' && req.params._id != req.user._id) return res.send(403);
```
pgarrison: Similarly, this should have a permission attached to it. `admin users` is the right one, I think?

public/angular/templates/incidents/index.html
```diff
@@ -21,7 +21,7 @@
         </div>
       </div>
     </div>
-    <div class="filter">
+    <div ng-hide="currentUser.role === 'viewer'" class="filter">
```
pgarrison: All these updates to the UI seem great, but can we attach it to permissions instead of roles?

pgarrison:
Since we've separated permissions from roles, we should use permissions instead of roles wherever possible, unless we decide this is over-engineered and we're only ever going to have these three roles.

lib/api/v1/user-controller.js
andresfib: Fixed. However, I have not been able to use Connect-Roles [methods](https://github.com/ForbesLindesay/connect-roles#requsercanaction-and-requserisaction) even if it should be [possible](https://github.com/jaredhanson/passport/wiki/Using-Passport-with-Connect-roles). Upgrading the package has not helped. Any idea, @hooverlunch ?

lib/api/v1/user-controller.js
```diff
@@ -2,14 +2,15 @@
 
 var express = require('express');
 var User = require('../../../models/user');
+var UserPermissions = require('../../../shared/user.js');
```
pgarrison: rather `'../../../shared/user`

pgarrison: rather `'../../../shared/user'`

lib/api/v1/user-controller.js
```diff
@@ -2,14 +2,15 @@
...
 var mailer = require('../../mailer');
 var _ = require('underscore');
 
 module.exports = function(app, user) {
   app = app || express();
-  user = user || require('../authorization')(app);
   var auth = require('../authentication')(app);
   var password = require('../reset-password')(app, auth);
+  user = user || require('../authorization')(app);
```
pgarrison: If it is significant that this gets `require`d after the others, there should be a comment saying so.

lib/api/v1/user-controller.js
```diff
@@ -34,15 +35,19 @@ module.exports = function(app, user) {
   app.use(express.json());
 
   // Get a list of all Users
-  app.get('/api/v1/user', user.can('manage users'), function(req, res) {
-    User.find({}, '-password', function(err, users) {
+  app.get('/api/v1/user', user.can('view users'), function(req, res) {
+    var query = {};
+    if (req.user && (new UserPermissions(req.user)).can('not view other users'))
```
pgarrison: looks good enough

lib/api/v1/user-controller.js
```diff
@@ -34,15 +35,19 @@ module.exports = function(app, user) {
...
+      query.username= req.user.username;
```
pgarrison: space before `=`

pgarrison:
Tests are passing. There are no tests here, but as you've said Andres, using `adminParty` for testing makes this super hard to test.

With the commit I just added there are no more references to the `manager` role. If you apply these updates with an existing Aggie database, the manager role breaks. An `admin` can still log in and change them to monitors, but the interface shows the manager role as blank.

andresfib:
Yes, this change in the database accepted values should be a good time to try using the migrate tool. I will come with a migration script soon.

pgarrison:
Okay, migration works, tests are passing (ignore travis).

lib/api/v1/user-controller.js
hooverlunch: I've never seen permissions defined in the negative like this. I think it's probably not a good idea. I assume this is done to handle the viewer role being able to see their own profile. But why would they ever need to list _all_ users? Could you just mark this one 'view users' and be done with it? Or if it's really needed, then I'd suggest 'view other users' and 'view profile' permissions.

lib/api/v1/user-controller.js
```diff
@@ -67,10 +72,12 @@ module.exports = function(app, user) {
   });
 
   // Update a User
-  app.put('/api/v1/user/:_id', user.can('manage users'), function(req, res) {
+  app.put('/api/v1/user/:_id', user.can('update users'), function(req, res) {
```
hooverlunch: Since all roles can update their own profile you perhaps don't need to protect this one at all, and just use your 'admin' check inside the action.

public/angular/js/services/auth.js
```diff
@@ -19,7 +20,7 @@ angular.module('Aggie')
           var promise = $http.get('/session');
           promise.then(function(res) {
             if (res.data.username) {
-              $rootScope.currentUser = res.data;
+              $rootScope.currentUser = new shared.User(res.data);
```
hooverlunch: Nice! Always good to see more sharing!

public/angular/templates/incidents/table.html
```diff
@@ -1,21 +1,21 @@
 <table class="table" ng-show="pagination.total > 0 || modal">
   <thead>
     <tr>
-      <th ng-hide="modal"></th>
+      <th ng-hide="modal || !currentUser.can('edit data')"></th>
       <th class="compact" translate>ID#</th>
       <th class="compact" translate>Title</th>
       <th class="compact" translate>Location</th>
-      <th class="compact" translate>Assigned To</th>
+      <th ng-hide="currentUser.can('not view other users')" class="compact" translate>Assigned To</th>
```
hooverlunch: Same here.

public/angular/templates/incidents/table.html
```diff
@@ -30,7 +30,7 @@
         <span ng-if="i.locationName">{{ i.locationName }}</span>
         <span ng-if="!i.locationName && i.latitude && i.longitude">{{ i.latitude | number:6 }}, {{ i.longitude | number:6 }}</span>
       </td>
-      <td class="compact"><a href="profile/{{i.assignedTo.username}}">{{ i.assignedTo.username }}</a></td>
+      <td ng-hide="currentUser.can('not view other users')" class="compact"><a href="profile/{{i.assignedTo.username}}">{{ i.assignedTo.username }}</a></td>
```
hooverlunch: Same here.

public/angular/templates/sources/index.html
```diff
@@ -11,10 +12,10 @@
     <tr>
       <th class="compact" translate>Media</th>
       <th class="compact" translate>Name</th>
-      <th class="compact" translate>Creator</th>
+      <th class="compact" ng-hide="currentUser.can('not view other users')" translate>Creator</th>
```
hooverlunch: Same here.

public/angular/templates/trends/bars.html
```diff
@@ -26,11 +26,11 @@
...
         <aggie-sparkline values="t.counts" start-time="startTime" height="t.minmax" callback="showReport(sparkEvent, t)"></aggie-sparkline>
       </td>
-      <td class="compact text-center status">
+      <td ng-show="currentUser.can('edit data')" class="compact text-center status">
         <aggie-toggle toggle="t.enabled" boolean options="{ true: 'On', false: 'Off' }" on-change="toggleEnabled(t)">
       </td>
-      <td class="compact text-center middle">
-        <button type="button" href class="btn btn-info" aggie-confirm="Are you sure you want to delete this trend? (All historical data will be lost)" on-confirm="deleteTrend(t)">{{ 'Delete' | translate }}</button>
+      <td ng-show="currentUser.can('edit data')" class="compact text-center middle">
+        <button type="button" href class="btn btn-info" aggie-confirm="{{ 'Are you sure you want to delete this trend? (All historical data will be lost)' | translate }}}" on-confirm="deleteTrend(t)">{{ 'Delete' | translate }}</button>
```
hooverlunch: Nice I18n catch here!

public/angular/templates/users/index.html
```diff
@@ -29,7 +29,7 @@
       <td class="compact content">{{ u.role | translate }}</td>
       <td ng-controller="UserFormModalController" class="actions compact">
         <a href ng-click="edit(u)"><i class="fa fa-pencil text-blue"></i></a>
-        <a href aggie-confirm="{{ 'Are you sure you want to delete this user?' | translate }}" on-confirm="delete(u)"><i class="fa fa-trash text-blue"></i></a>
+        <a href ng-show="currentUser.role === 'admin'" aggie-confirm="{{ 'Are you sure you want to delete this user?' | translate }}" on-confirm="delete(u)"><i class="fa fa-trash text-blue"></i></a>
```
hooverlunch: Why not using `can` here?

public/angular/translations/locale-en.json
```diff
@@ -14,7 +14,6 @@
   "unassigned": "Unassigned",
   "viewer": "Viewer",
   "monitor": "Monitor",
-  "manager": "Manager",
```
hooverlunch: Wow very thorough, _removing_ i18n strings!

shared/user.js
```diff
@@ -11,13 +11,13 @@
   User.PASSWORD_MIN_LENGTH = 8;
 
   User.permissions = {
-    'view data': ['viewer', 'monitor', 'manager', 'admin'],
-    'edit incidents': ['monitor', 'manager', 'admin'],
-    'edit reports': ['monitor', 'manager', 'admin'],
-    'edit sources': ['manager', 'admin'],
-    'toggle fetching': ['manager', 'admin'],
-    'manage trends': ['manager', 'admin'],
+    'view data': ['viewer', 'monitor', 'admin'],
```
hooverlunch: If we were doing this over again I would have suggested not removing the granularity (incidents/reports/sources) in favour of a more generic 'data', because it handcuffs us down the line if we want to tweak permissions. If you think that could be reinstated without too much effort I would do it, but this is probably okay also for now.

hooverlunch:
Looking really nice folks. This is mostly done according to best practices. Just one weirdness with a negative permission that I think should be refactored. A couple other smaller things as well, but I don't need to see it again.

Thanks!

PS Please make sure Andy has & knows about the permissions table. It should go in the docs.

lib/api/v1/user-controller.js
andresfib: I agree with the negative not being a great idea, I will make it positive. 

In my understading for privacy and authorization viewers shouldn't be able to list all users. However, in the current implementation the incidents, and others, angular controller does require User.query() to grab all the users for things like filtering. See [routes.js](https://github.com/TID-Lab/aggie/blob/651b37318160e656d9dc307b5b27fa819aa6dce9/public/angular/js/routes.js#L130). That's the reason why everybody has access to this endpoint. 

I couldn't think of a better way to get around that requirement.

pgarrison:
It seems the permissions aren't working correctly now. As a viewer, I can see all users, turn admins into viewers, and add new admins.

pgarrison:
Looks like all appropriate changes have been made. Automatic tests pass and most tests in the `testing.md` document pass (the ones that don't are not regressions). I'll rebase and merge.

andresfib:
Check admin party...

pgarrison:
Deleted that comment because I realized I had switched back to the develop branch to test something!

PR #168, opened 2016-06-17
4597: Fixed bug

andresfib:
Solution here:
https://github.com/wpalahnuk/ngAutocomplete/issues/62


pgarrison:
Code appears fine, but I'm not sure what the bug was so I can't really tell if it's fixed. Write a test?

pgarrison:
Ah, I found at least one difference. a google maps pinpoint icon shows up next to an incident's location coordinates.

pgarrison:
Okay, figured it out, looks good to me. My adblocker thought our maps API calls were ads.

PR #169, opened 2016-07-14
4599 add incident from report page

andresfib:
Now Adding an incident to a single or multiple report works better. Incident can be a new one or an old one. Newly created incidents are fetched so that linked reports show the incident title without the need for a browser refresh.

This also fixes Bug #4599: Should not be redirecting to incident page when creating incident via report page


pgarrison:
Tests pass, code looks good, seems like it works.

pgarrison:
Rebased onto develop to accommodate #168 and noticed that the problem from #168 applies to incidents created from the reports page still. New bug 4800 covers this.

PR #170, opened 2016-07-15
4797: Fixed raced condition when increasing totalReports

andresfib:
Linking reports to incident fired consecutives db finds to the same incident which updated the values in the callback. Now, we use atomic call findByIdAndUpdate.

Added test to check fix.

### DIFF
--- a/models/incident.js
+++ b/models/incident.js
@@ -70,14 +70,11 @@ schema.plugin(autoIncrement.plugin, { model: 'Incident', field: 'idnum', startAt
 
 schema.listenTo(Report, 'change:incident', function(prevIncident, newIncident) {
   if (prevIncident !== newIncident) {
-    Incident.findByIdAndUpdate(prevIncident, { $inc: { totalReports: -1 } }, function(err, incident) {
-      if (err || !incident) return;
-    });
+    //Callbacks added to execute query immediately
+    Incident.findByIdAndUpdate(prevIncident, { $inc: { totalReports: -1 } }, function(err, incident) {});
   }
 
-  Incident.findByIdAndUpdate(newIncident, { $inc: { totalReports: 1 } }, function(err, incident) {
-    if (err || !incident) return;
-  });
+  Incident.findByIdAndUpdate(newIncident, { $inc: { totalReports: 1 } }, function(err, incident) {});
 
 });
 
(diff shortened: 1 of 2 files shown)


models/incident.js
```diff
@@ -69,20 +69,16 @@ var Incident = mongoose.model('Incident', schema);
...
-    incident.totalReports = total;
+  if (prevIncident !== newIncident) {
+    Incident.findByIdAndUpdate(prevIncident, { $inc: { totalReports: -1 } }, function(err, incident) {
+      if (err || !incident) return;
+    });
+  }
 
-    incident.save();
+  Incident.findByIdAndUpdate(newIncident, { $inc: { totalReports: 1 } }, function(err, incident) {
+    if (err || !incident) return;
```
pgarrison: This callback does nothing. Better to use _.identity and a comment, or better yet, handle errors. Similarly above.

test/lib.api.v1.report-controller.test.js
```diff
@@ -183,18 +200,31 @@ describe('Report controller', function() {
...
             .end(function(err, res) {
               expect(res.body).to.have.property('_incident');
-              expect(res.body._incident).to.equal('54c73024ae04d1f9c3a678d6');
+              expect(res.body._incident).to.equal(String(incidents[0]._id));
               done();
             });
           });
         });
     });
+    it('should update the totalReports field in incident', function(done) {
```
pgarrison: Tiny change, but empty line above this?

test/lib.api.v1.report-controller.test.js
```diff
@@ -183,18 +200,31 @@ describe('Report controller', function() {
...
+      request(reportController)
+        .patch('/api/v1/report/_link')
+        .send({ ids: [reports[2].id], incident: incidents[0]._id })
+        .expect(200)
+        .end(function(err) {
+          if (err) return done(err);
+          Incident.findById(incidents[0]._id, function(err, incident) {
+            expect(incident.totalReports).to.equal(3);
+            done();
```
pgarrison: `done(err);`?

pgarrison: This test would be even better if we first check `expect(incident.totalReports).to.equal(2);` or (extra nice) if we then remove the report and check that it goes down to 2.

pgarrison:
To replicate the bug this fixes in the browser:
* make an incident
* from the reports page, add more than one report to that incident at once (with the checkboxes)
* go to the page of all the incidents
* in the Title column, below the title of this incident the number of reports is too low (just 1)

pgarrison:
Other than my comments there, the code looks great and this works.

pgarrison:
Rebased. Tests still pass.

PR #171, opened 2016-07-18
4602 source errors not disabling

andresfib:
Previously, source errors only disabled the source in the database. Now, the fetching process also listens to its own sources for a disable event, stopping the associated bot in that case


pgarrison:
Had some thoughts about what's going on here. What if we did this instead: de93b1408d

Why are we even listening to the `schema` from the `api` process? Shouldn't the one from the `fetching` process do just as well?

PR #173, opened 2016-08-02
Fix malformed string

pgarrison:
To reproduce the bug this fixes:

1. From reports page click add in the incident column [modal should
appear]


pgarrison:
(was issue 4902)

PR #175, opened 2016-08-05
Bugfix for change password button

pgarrison:
Summary: Subtly poor coding practices from long ago became a bug when
internationalization was added.

To reproduce the bug this fixes, go to `/users`, click the pencil by
your user, and click the Change Password button. This button wasn't
doing anything.

The cause was that adding the `translate` directive to the element
creates a new child scope. When the `ng-click` directive tried to update
the `showPassword` model it was only changing on the child scope. Best
practice is for `$scope` to always hold a reference to an object that
holds the primitives that make up the model. Then the object can be
shared between parent and child scopes. See
https://github.com/angular/angular.js/wiki/Understanding-Scopes

I read through the rest of the internationalization changes and it
seems this is the only place this comes up.

### DIFF
--- a/public/angular/templates/users/modal.html
+++ b/public/angular/templates/users/modal.html
@@ -62,10 +62,10 @@
             </div>
           </div>
           <div ng-show="user._id == currentUser._id">
-            <div class="form-group" ng-hide="showPassword || form.password.$dirty || form.passwordConfirmation.$dirty">
-              <button type="button" class="btn col-lg-6 col-lg-offset-3" ng-click="showPassword = !showPassword" translate>Change password</button>
+            <div class="form-group" ng-hide="model.showPassword || form.password.$dirty || form.passwordConfirmation.$dirty">
+              <button type="button" class="btn col-lg-6 col-lg-offset-3" ng-click="model.showPassword = true" translate>Change password</button>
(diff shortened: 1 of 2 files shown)


public/angular/templates/users/modal.html
```diff
@@ -62,10 +62,10 @@ <h4 class="modal-title">
             </div>
           </div>
           <div ng-show="user._id == currentUser._id">
-            <div class="form-group" ng-hide="showPassword || form.password.$dirty || form.passwordConfirmation.$dirty">
-              <button type="button" class="btn col-lg-6 col-lg-offset-3" ng-click="showPassword = !showPassword" translate>Change password</button>
+            <div class="form-group" ng-hide="model.showPassword || form.password.$dirty || form.passwordConfirmation.$dirty">
+              <button type="button" class="btn col-lg-6 col-lg-offset-3" ng-click="model.showPassword = true" translate>Change password</button>
```
andresfib: Maybe just ng-click="model.showPassword"

pgarrison: That wouldn't do anything; ng-click takes an expression to evaluate. I suppose we could write a showPassword function.

PR #180, opened 2016-08-30
4943 and 4939 Add button appearing twice and report redirection.

pgarrison:
PR #169 for 4599 removed the alternate "Add", and then PR #167 for 1785 added it back in improperly.

This fix leaves a bit to be desired, but it is very small and solves both of these bugs. Optimally, we would still use the `IncidentFormModalController` when there are no incidents yet. Adapting this modal for the reports page would take some extra work, however.


pgarrison:
Tests pass. Bugs appear fixed.

PR #182, opened 2016-09-01
4931 frontend testing

pgarrison:
Some basic frontend tests. Thanks to mkuklis for starting this a few years ago.

Introduces a defect that jquery is now checked into source. If we upgrade jquery to >=1.5.0, we can get it from npm instead.


andresfib:
Works  great in my computer!
In a way the first tests are quite similar to what we can get with mocha, but I can't wait to get more complicated ones with more steps.

PR #187, opened 2016-09-13
5082 setup protractor

pgarrison:
Install protractor, and reorganize tests. Test directory now looks like

```
test
├── backend
│   └── fixtures
├── end-to-end
├── frontend
│   ├── controllers
│   ├── services
│   └── vendor
└── manual
    └── logs
```

Run protractor with `npm run protractor` while also running Aggie with `npm run testrun`; these both connect to the `aggie-test` database and aggressively delete from there. `npm test` will still just run mocha and karma.

Our protractor setup uses mocha, instead of the default jasmine.

We have one end-to-end test in `test/end-to-end/login.test.js`, which logs the admin in for the first time, sets a new admin password, and then logs out and in again with the new password.

This also fixes a silly defect with dd082d7.


pgarrison:
codeclimate complains a bunch when you move files that have issues in them. In fact, there are just two new issues, both with 'npm-shrinkwrap.json`

```
"ws": {
          "version": "0.8.1",
          "from": "ws@>=0.8.0 <0.9.0",
          "resolved": "https://registry.npmjs.org/ws/-/ws-0.8.1.tgz"
        },
```

New issues:
```
ws DoS due to excessively large websocket message
ws Remote Memory Disclosure
```
These are in fact already present, as we depend on `ws ^1.0.1` already which has the same vulnerabilities. This outdated package is only being run by protractor, so there's no real risk here. Sadly, upgrading selenium (where this dependency is coming from) would require going up to protractor 3 or 4, which would mean a new version of node :'(

pgarrison:
All tests in `test` are passing.

README.md
```diff
@@ -56,13 +56,16 @@ The following need to be installed.
...
       1. in your terminal, navigate to the aggie project directory: `cd [aggie]`.
-      2. use this command: `nvm use` to install the version specified in `.nvmrc`.
+      2. use this command: `nvm use 0.10` to install the version specified in `.nvmrc`.
 2. **Mongo DB** (requires >= 2.6, such as 2.6.9)
   1. Follow the [installation structions](https://docs.mongodb.org/v2.6/) for your operating system.
   2. Stop and restart Mongo DB.
     1. On Linux run `sudo service mongod stop`. Then run `sudo mongod`.
     2. Make sure mongodb is running in the terminal and listening on an appropriate port. Your terminal with the mongo db process running should display something similar to the following: `[initandlisten] waiting for connections on port 27017`.
   3. Note: You do not need to create a user or database for aggie in Mongo DB. These will be generated during the installation process below.
+3. **JDK**
```
andresfib: I have tested it with JRE and it works. Thus, we may prefer asking people to install JRE instead of the whole JDK.

pgarrison:
Updating the `gulpfile` is a good idea, at least for karma. I realize running tests with gulp is broken since I moved the tests. Since protractor requires that you also be running aggie on the test database in the background, we should think more about what we want. Possible issues are (1) you've got another aggie instance running (2) we might not want to start aggie through gulp and (3) you update files more quickly than the end-to-end tests run.

PR #189, opened 2016-09-23
5052 filtered batch

pgarrison:
If you set a filter on the reports page and then click 'Grab Batch', it will only give you reports that match your filter.

There's still some behavior that is possibly unintuitive. We might want to think about what this should be. Here are a couple possibly weird situations

1. Someone sets a filter like "status: read, author: philip" on the reports page. They'll only see things that are marked read. Then when they click grab batch, they'll get _unread_ reports with "author: philip". Should grab batch just error here with "No more unread reports found"?
2. Right now, the only way to tell that your batch is not a normal unfiltered batch is by looking at the URL bar. If you don't realize you have a filter set, this could be confusing.
3. If you check out a batch with a filter and then navigate to `reports/batch` with no query string, your next batch (through mark all and grab another) will not apply the filter that you had. I'm not convinced this is a defect, but again it could be confusing.
4. If you check out a batch from page `n` of the reports page, when you return to the reports page via e.g. the 'Cancel' button, your filter will still be applied but it will take you to page `1`. I'm not sure which page is best to return to.

This feature comes with a couple basic backend unit tests. The interface there hasn't really changed, but now `Batch.checkout` and `Batch.load` take a `query` parameter. It also comes with a bunch of neat end-to-end tests with protractor, and some manual end-to-end tests in `testing.md`. Except for a manual end-to-end test that catches issue 5147 (which was already there), all of these pass on my machine.

For the future implementation of tags, when we actually save filters to the database, we should probably make a `models/filter.js` and in there convert between the canonical filter object it defines and what gets passed from `$scope.searchParams`.

These changes get us pretty close to fixing issue 1927, and I've filed three newly found issues related to this PR: 

- 5146: ReportsIndexController wants to be refactored, since I've add a little more cruft to this already unweildy code
- 5147: Adding report to incident from batch doesn't update table, which is a major (preexisting) bug affecting the grab batch feature
- 5152: Potentially confusing behavior of grab batch with filter, to document the four situations I described above


pgarrison:
The issues code climate found are not actually new -- just lines I touched (and probably fixed some style issues on).

testing.md
```diff
@@ -64,3 +64,22 @@ Users and emails
...
+3. Click an incident [modal closes, text `Add` is updated to the name of the
+incident]
+4. Click `Mark All and Done` [goes back to /reports and the reports from the
+batch are marked as read]
+5. Navigate to `/reports/batch` [no reports are displayed]
+6. Navigate to `/reports` and click `Grab Batch` again [goes to `/reports/batch`
+and populates with some reports]
+7. Log in as a different user in a different tab/window
+8. Navigate to `/reports` as the second user and click `Grab Batch`. [reports
+are populated, and none of them are reports the other user has in their batch]
```
andresfib: Maybe add one test for Filtered Smart Batch once we have the UI nailed down.

pgarrison: I had some, and then I automated them!

pgarrison:
I've responded to the comments with the latest commit (49383 Mention filters applied to a batch, badly) and filed issue 5311 to document the parts of that I find unsavory.

PR #190, opened 2016-09-26
Fixes duplication of reports and other events

karthiksbhat:
Events like report generation, bot creation etc were being duplicated because of two listeners set up to listen to the same events. There is still possibility of discovering a bug related to this in the future: there is a listener for all events emitted by the API process, but only for 1 event emitted by the Fetching process. But currently, this code fixes the known bugs related to this code. The changes to bot-master.js have been made to mimic Andres' PR (PR 171).
End to end tests: 
Some changes have been copied from Philip's other branch to avoid future merge conflicts.
Toolkit has some new useful functions which can be reused in other tests.
duplication-test.js: This is a reusable test which will add new source, listen for reports, and check if the correct number of reports are generated. This was designed to check duplication of reports.

### DIFF
--- a/lib/fetching/bot-master.js
+++ b/lib/fetching/bot-master.js
@@ -164,8 +164,8 @@ BotMaster.prototype._loadAll = function(callback) {
   // Find sources from the database
   Source.find({ media: { '$ne': null } }, function(err, sources) {
     if (err) return callback(err);
-    async.each(sources, function(s,c){
-       self._loadOne(s,c); 
+    async.each(sources, function(s, c) {
+       self._loadOne(s, c);
     }, callback);
   });
 };
--- a/lib/fetching/bot-master.js
+++ b/lib/fetching/bot-master.js
@@ -1,5 +1,6 @@
 // Creates and destroys bots as sources are created, enabled, disabled, destroyed, etc.
 // Watches for fetching to be turned on and off and disables/enables bots accordingly.
+'use strict';
 
 var _ = require('underscore');
 var Source = require('../../models/source');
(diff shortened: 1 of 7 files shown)


test/end-to-end/duplication.test.js
```diff
@@ -0,0 +1,166 @@
...
+      utils.addSource('SMS GH', { nickname: 'hello', keywords: 'test' });
+    })
+    .then(function() {
+      utils.toggleFetching('On');
+    })
+    .then(function() {
+      utils.toggleSource('SMS GH', 'On');
+    })
+    .then(function() {
+      request('http://localhost:1111')
```
andresfib: make a function of the SMS Gh request. You can also abstract the 5 test, so you only pass the status of the toggles and the expected reports to a function, and don't duplicate so much code

lib/fetching/bot-master.js
```diff
@@ -107,15 +109,29 @@ BotMaster.prototype._addSourceListeners = function(emitter) {
 
   // Listen to `enable` event from the API process
   emitter.on('source:enable', function(source) {
+    // This should ensure fetching button does its job
```
andresfib: To a different PR

lib/fetching/bot-master.js
```diff
@@ -107,15 +109,29 @@ BotMaster.prototype._addSourceListeners = function(emitter) {
...
+    if (!self.enabled) {
+      return;
+    }
     var bot = self._getBot(source._id);
     if (self.enabled && bot) bot.start();
   });
 
   // Listen to `disable` event from the API process
   emitter.on('source:disable', function(source) {
+    // This should ensure fetching button does its job
```
andresfib: To a different PR

karthiksbhat:
Closing because the issues handled by this PR have been split and submitted as separate PR's, as suggested.

PR #191, opened 2016-09-27
5043 Duplication bug is fixed

karthiksbhat:
This contains a fix for the duplication of reports that was occurring in Twitter and SMSGh (noticed). There were two listeners for the same events, leading to duplication. There is still possibility of discovering a bug related to this in the future: there is a listener for all events emitted by the API process, but only for 1 event emitted by the Fetching process. But currently, this code fixes the known bugs related to this code. The changes to bot-master.js have been made to mimic Andres' PR (PR 171).
This also contains an end-to-end test to check how Aggie reacts to different states of Fetching and Sources.

### DIFF
--- a/test/end-to-end/source-fetching-toggle.test.js
+++ b/test/end-to-end/source-fetching-toggle.test.js
@@ -28,19 +28,6 @@ describe('test duplication of reports with different settings', function() {
     defer.fulfill(true);
     return defer.promise;
   };
-
-  var sendRequest = function() {
-    request('http://localhost:1111')
-    .get('/smsghana')
-    .query(reqParams)
-    .expect(200)
-    .end(function(err, res) {
-      if (err) {
-        return done(err);
-      }
-    });
-  };
-
   it('should listen with fetching:on and source:enabled', function(done) {
     chain()
     .then(function() {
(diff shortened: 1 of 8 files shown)


test/end-to-end/source-fetching-toggle.test.js
```diff
@@ -0,0 +1,145 @@
...
+      sendRequest();
+    });
+    browser.get(browser.baseUrl + 'reports');
+    expect(utils.getReports().count()).to.eventually.equal(1);
+    utils.toggleSource('SMS GH', 'Off');
+    utils.toggleFetching('Off');
+    utils.deleteSource('SMS GH', 'hello');
+    done();
+  });
+  it('should not listen with fetching:on and source:disabled', function(done) {
```
andresfib: Add a blank line here to be consistent

test/end-to-end/source-fetching-toggle.test.js
```diff
@@ -0,0 +1,145 @@
+'use strict';
```
andresfib: Indentation is not correct in several places, e.g.,  `var reqParams =...` File ends with two closing braces on the same column.

pgarrison:
These tests won't fail correctly. Here's an idea of how it would (I think) work. https://gist.github.com/pgarrison/65077206bbf483680b286f0885170126

pgarrison:
The tests here seem very useful. `should not listen with fetching:on and source:disabled`, if it were working, would I think catch 5102. Unfortunately these tests are passing when they should be failing or are failing for the wrong reason. Maybe we can fix this PR by removing some of the tests that apply to other issues (5102), and then reintroduce those tests elsewhere?

pgarrison:
Okay, documented this in 5275

PR #194, opened 2016-10-05
4800 incident cords modal

andresfib:
Admin can now set the key in the settings page. The key is added to the
script request to Google servers.
Key can be retrieved by anyone, so that library is loaded even when not logged in. 
Refactored Settings to accommodate for the change.

Still using the same ng-Autocomplete library as in the beginning


andresfib:
Code climate new errors because of moving file around!

pgarrison:
Andres can you explain this comment to me `Get Google Places API Key. Remember to set it so that only certain domains can use it`?

andresfib:
Everyone accessing the Aggie deployment will be able to see the key in the source code of index.html. People should whitelist the addresses that can use the key so 'unathorized' requests are not taken into account by Google. By signing up we get 1K free requests per day, by adding a credit card we get 150K.
https://developers.google.com/places/web-service/usage

This may go in the README...

PR #195, opened 2016-10-05
5102 fetching reliability

karthiksbhat:
The fetching button was unreliable and when had issues like fetching even when disabled and vice versa. Also, if a source was disabled before fetching, it would be force-enabled when fetching was toggled from off to on.
In this PR, the fetching button has been modified to send events only when toggled. (previously, fetching:start and fetching:stop events were being sent if the button was clicked even if it was already in that state). Next, each time fetching is started, the bot-master checks the database for enabled bots, and starts only those bots. By doing so, fetching ensures that there is no force-enabling of bots. This also maintains a consistency between the state of the bot in the sources page, and the actual state of the bot.

### DIFF
--- a/lib/fetching/bot-master.js
+++ b/lib/fetching/bot-master.js
@@ -44,7 +44,7 @@ BotMaster.prototype.start = function() {
   var self = this;
   this.isFetching = true;
   this.bots.forEach(function(bot) {
-    self.enableBot(bot, function() {});
+    self.enableBot(bot);
   });
 };
 
@@ -53,7 +53,7 @@ BotMaster.prototype.stop = function() {
   var self = this;
   this.isFetching = false;
   this.bots.forEach(function(bot) {
-    self.disableBot(bot, function() {});
+    self.disableBot(bot);
   });
 };
 
@@ -63,14 +63,14 @@ BotMaster.prototype.stop = function() {
 BotMaster.prototype.kill = function(bot) {
   if (bot) {
     // Stop bot to ensure unbinding of listeners
-    this.disableBot(bot, function() {});
+    this.disableBot(bot);
     // Remove bot instance from list of bots
     var index = this.bots.indexOf(bot);
     if (index > -1) this.bots.splice(index, 1);
   } else {
     // Kill all bots
     for (var i in this.bots) {
-      this.disableBot(this.bots[i], function() {});
+      this.disableBot(this.bots[i]);
       this.bots[i] = null;
       delete this.bots[i];
     }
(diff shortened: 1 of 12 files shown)


lib/fetching/bot-master.js
```diff
@@ -42,18 +42,18 @@ BotMaster.prototype.reset = function() {
 // Start all bots
 BotMaster.prototype.start = function() {
   var self = this;
-  this.enabled = true;
+  this.isFetching = true;
   this.bots.forEach(function(bot) {
-    self.enableBot(bot);
+    self.enableBot(bot, function() {});
```
andresfib: No need for the empty function. Just because you feel like the Source.get requires a callback(err), you don't need the empty function. Changing the enable and disable function signatures is what probably has broken the tests.

lib/fetching/bot-master.js
```diff
@@ -63,31 +63,42 @@ BotMaster.prototype.stop = function() {
...
+// This is called when a new bot is added, and when botmaster.start is called.
 // Enable bot from current process
-BotMaster.prototype.enableBot = function(bot) {
+BotMaster.prototype.enableBot = function(bot, callback) {
   // Bail if fetching is disabled
-  if (!this.enabled) return;
-  if (bot.source.enabled) bot.start();
+  if (!this.isFetching) return;
+  Source.findById(bot.source._id, function(err, source) {
+    if (err) return callback(err);
```
andresfib: Don't callback. You could, for example, just log a warning that bot could not be enabled because its source does not exist, and print the error message.

lib/fetching/bot-master.js
karthiksbhat: Possibly. Working on it now.

pgarrison:
It appears that the reason the tests are failing now is that `BotMaster.start` is asynchronous but doesn't take a callback. The tests call `botMaster.start();` but then proceed immediately.

lib/fetching/bot.js
```diff
@@ -34,14 +34,15 @@ util.inherits(Bot, EventEmitter);
 Bot.prototype.start = function() {
   if (this.enabled) return;
   this.enabled = true;
+  this.source.enabled = true;
```
pgarrison: This line worries me. Seems like we're breaking an abstraction if a bot can modify its source (especially without checking in with the db!)

karthiksbhat: The major problem was that the status of the source was not being properly reflected in the bot-master. So, at some point, the source.enabled would be 'true', but the bot would've been killed (by disabling the source on the sources page). In this case, when Fetching was toggled, it somehow used to re-enable a source, despite the sources page saying it was disabled.

pgarrison:
This fixes the tests in #203

andresfib:
All tests passing in my machine. Great work on creating and cleaning up the tests.

Now let's look forward for a more stable future!

PR #196, opened 2016-10-06
4496: Specific facebook error codes emit warnings rather than errors

andresfib:
A list of transient errors has been created from https://developers.facebook.com/docs/graph-api/using-graph-api/#errors. Experience may lead us to change that list as time evolves.

Comes with a test!

### DIFF
--- a/lib/fetching/content-services/facebook-content-service.js
+++ b/lib/fetching/content-services/facebook-content-service.js
@@ -41,6 +41,9 @@ var graph = require('fbgraph');
 var util = require('util');
 var _ = require('underscore');
 
+// From https://developers.facebook.com/docs/graph-api/using-graph-api/#errors
+var transientErrorCodes = [1, 2, 4, 17, 341];
+
 // options.url - The url of the Facebook resource
 // options.lastReportDate - The time of the last report already fetched (optional)
 var FacebookContentService = function(options) {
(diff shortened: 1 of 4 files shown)


lib/fetching/content-services/facebook-content-service.js
```diff
@@ -82,9 +85,21 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
     return;
   }
 
+  var isTransient = function(error) {
+    if (error.code && _.contains(transientErrorCodes, error.code)) {
+      return true;
+    }
+    return false;
```
pgarrison: `return error.code && _.contains(transientErrorCodes, error.code);`

lib/fetching/content-services/facebook-content-service.js
```diff
@@ -82,9 +85,21 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
+  };
+
   this._doRequest(source, function(err, data) {
     if (err) {
-      self.emit('error', new Error(err.message));
+      var level = 'error';
+      if (isTransient(err)) level = 'warning';
+      process.nextTick(function() {
```
pgarrison: Any reason this needs to be async?

pgarrison:
Codeclimate has some reasonable complaints here

PR #200, opened 2016-10-12
4802: Added error handling for Incident update

karthiksbhat:
Normal usage does not seem to trigger this error. So, if triggered, it is something serious.
Logger will, therefore, report it as an error.


pgarrison:
We can fix this in a bunch of other places too. See
```
eslint --no-eslintrc --rule '{handle-callback-err: 2}' lib models
```
(and maybe some relevant ones in other folders)

PR #203, opened 2016-10-14
5102 A couple tests for the fetching toggling issue

pgarrison:
These tests require access to the Twitter API, so we've added a new
command `npm run protractor-with-apis` which is documented in
`test/end-to-end/with-apis/README.md`.

Four tests are added here. The first two are passing, and the second two
are failing.

These tests all pass on the current 5102_fetching_reliability:
```
git checkout 4544ab9c2
git cherry-pick 6f5c896
npm i
npm run testrun &
npm run protractor-with-apis
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ The following need to be installed.
 4. To run end-to-end tests:
   1. first start Aggie on the test database with `npm run testrun`
   2. then run protractor with `npm run protractor`
+5. To run end-to-end tests with external APIs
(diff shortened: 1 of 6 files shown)


npm-debug.log.377800843
```diff
@@ -0,0 +1,24 @@
+0 info it worked if it ends with ok
```
andresfib: maybe we dont want this in github

PR #204, opened 2016-10-24
1927 remove Mongo text search

pgarrison:
The very old `mongoose-text-search` plugin was using deprecated functionality in mongo. Now it's gone and we're closer to being able to upgrade mongo.

This is created from the 5052 branch, which took the first step of removing some of our calls to the text search plugin.

### DIFF
--- a/lib/analytics/trend-queryer.js
+++ b/lib/analytics/trend-queryer.js
@@ -4,14 +4,17 @@
 var Trend = require('../../models/trend');
 var ReportQuery = require('../../models/query/report-query');
 var Report = require('../../models/report');
-var _ = require('underscore');
+var _ = require('lodash');
+var logger = require('../logger');
 
 var TrendQueryer = function(options) {
   this.trend = options.trend;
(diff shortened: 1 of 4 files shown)


lib/analytics/trend-queryer.js
```diff
@@ -112,48 +94,33 @@ TrendQueryer.prototype.backFill = function(callback) {
...
-    this.runReportAnalysis(query.keywords, options, function(err, trends) {
-      if (err) return callback(err);
-      if (!trends.length) return callback(null, []);
-      // Sort by timebox
-      trends = _.sortBy(trends, 'timebox');
-      // Store trend counts in trend model
-      self.updateTrends(trends, callback);
-    });
-  }
+  // Just use filters when no keywords are provided
```
andresfib: Comment confusing! Looks like you are using keywords in the filter.

pgarrison: Good catch!

PR #205, opened 2016-10-24
4803: Sources switch back to 'Off' after every warning

karthiksbhat:
When a bot is stopped, it alters the 'enabled' flag in its source object. This (somehow) is accessed by the sources in the 'Sources' page, and when a source errors and its bot is stopped, the Source state toggles to 'off'.


karthiksbhat:
Handled in a different (better) way by #207 .

PR #207, opened 2016-10-25
4803 When sources error, the fetching button turns off

pgarrison:
The source schema in the fetching process gets out of date when clicking
'On' or 'Off' from `/sources`. Only the `source:save` event triggers
reloading the source, not `source:enable` or `source:disable`. Now we
emit `source:save` whenever a source is saved.

Two other changes here that I ran into
 * in `bot.js`, `logEvent` sends its errors to the
   callback, which should not be ignored.
 * `$scope.refresh` from `SourcesIndexController` is not used anywhere


pgarrison:
We should probably write a test for this

pgarrison:
All protractor tests are passing.

PR #210, opened 2016-10-26
5329 Add a tag field to the incident filters

pgarrison:
Syling the two (or more) rows of the incident filter now uses flexbox.
This does not support IE 8.

Since much of this styling code is rightly shared by the report filter,
flexbox now controls that too. Some elements in the filters (both on
reports and incidents) are now allowed to be their full size (instead of
cutting off text), which means that on not-super-small screens, the
filter on the reports page overflows to two rows.

This does open us up to one user-created sort-of issue: if a user makes an
incident or source with a really long name, then the filter element for
slecting an incident/source will be correspondingly large. Source and
incident names are limited to 20 and 32 characters each, so this
shouldn't really be a big problem now.


pgarrison:
P.S. it's probably best to view these changes with `git diff -w -M`

pgarrison:
Style issues here are not new, just worse :/

pgarrison:
Also passes tags field to `GET /api/v1/incident/:id`.

PR #212, opened 2016-10-28
5338 frontend tag processing

pgarrison:
Frontend logic for filtering incidents by tag. Passes `tags` field to `GET /api/v1/incident/:id`.

Depends on #210.


pgarrison:
Style issues here are not new, just worse :/

PR #213, opened 2016-10-28
5327 add tag to incident

pgarrison:
The incident modal now has a field to add a list of tags to an incident (id is `tags`), and the table at `/incidents` has a column for an incident's tags. Also resolves 5328.


karthiksbhat:
Looks good to merge.

PR #221, opened 2016-11-16
5403 edit incident

pgarrison:
Editing incidents (without editing the tags) was timing out. Fix was to rollback some changes from 4e0fe80 (#215). Added an e2e test that breaks without the fix. 

Also made a fix to a major issue with creating incidents in e2e tests (see diff of `test/database-tools.js`).


karthiksbhat:
Looks good!

PR #222, opened 2016-11-17
5356 logging

karthiksbhat:
Log user interaction with Aggie. Storing actions like:
Reports read/flagged
Report added to incident
Incident created/edited/viewed
Source created/edited/deleted

### DIFF
--- a/config/secrets.json.example
+++ b/config/secrets.json.example
@@ -55,6 +55,7 @@
     "api": {
       "log_requests": true,
       "log_responses": true,
+      "log_user_activity": false,
       "filename": "logs/api.log"
     },
     "master": {
--- a/lib/api/v1/incident-controller.js
+++ b/lib/api/v1/incident-controller.js
@@ -19,7 +19,16 @@ module.exports = function(app, user) {
       if (err) {
         res.send(err.status, err.message);
       } else {
-        logger.logActivity();
+        var logDetails = {
+          userID: req.user._id,
+          username: req.user.username,
+          action: 'create',
+          actionRef: {
+            incidentTitle: incident.title,
+            incidentID: incident._id
+          }
+        };
+        logger.logActivity(logDetails);
         res.send(200, incident);
       }
     });
@@ -46,7 +55,19 @@ module.exports = function(app, user) {
       .exec(function(err, incident) {
         if (err) res.send(err.status, err.message);
         else if (!incident) res.send(404);
-        else res.send(200, incident);
+        else {
+          var logDetails = {
+            userID: req.user._id,
+            username: req.user.username,
+            action: 'view',
+            actionRef: {
+              incidentTitle: incident.title,
+              incidentID: incident._id
+            }
+          };
+          logger.logActivity(logDetails);
+          res.send(200, incident);
+        }
       }
     );
   });
@@ -67,7 +88,16 @@ module.exports = function(app, user) {
         } else if (!numberAffected) {
           res.send(404);
         } else {
-          logger.logActivity();
+          var logDetails = {
+            userID: req.user._id,
+            username: req.user.username,
+            action: 'edit',
+            actionRef: {
+              incidentTitle: incident.title,
+              incidentID: incident._id
+            }
+          };
+          logger.logActivity(logDetails);
           res.send(200);
         }
       });
@@ -87,8 +117,17 @@ module.exports = function(app, user) {
         // Delete each incident explicitly to catch it in model
         incident.remove(function(err) {
           if (err) return res.send(err.status, err.message);
+          var logDetails = {
+            userID: req.user._id,
+            username: req.user.username,
+            action: 'delete',
+            actionRef: {
+              incidentTitle: incident.title,
+              incidentID: incident._id
+            }
+          };
+          logger.logActivity(logDetails);
           if (--remaining === 0) {
-            logger.logActivity();
             return res.send(200);
           }
         });
@@ -104,9 +143,18 @@ module.exports = function(app, user) {
       if (!incident) return res.send(404);
       incident.remove(function(err) {
         if (err) {
-          logger.logActivity();
           return res.send(err.status, err.message);
         }
+        var logDetails = {
+          userID: req.user._id,
+          username: req.user.username,
+          action: 'delete',
+          actionRef: {
+            incidentTitle: incident.title,
+            incidentID: incident._id
+          }
+        };
+        logger.logActivity(logDetails);
         res.send(200);
       });
     });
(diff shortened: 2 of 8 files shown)


lib/api/v1/incident-controller.js
```diff
@@ -109,6 +139,8 @@ module.exports = function(app, user) {
         // Delete each incident explicitly to catch it in model
         incident.remove(function(err) {
           if (err) return res.send(err.status, err.message);
+          var logDetails = writelog.writeIncident(req, incident, 'deleteIncident');
```
andresfib: Could write report be a function insider the logger  library? We just pass the parameters to logger.logActivity

karthiksbhat: I don't want to modify the structure of logger too much. I have tried to avoid logger from accessing the database (therefore creating userLogger.js), or in this case, from accessing secrets.json.
What I did was call logger from writeLog, so there's no unnecessary returning.

lib/api/v1/incident-controller.js
```diff
@@ -34,14 +40,20 @@ module.exports = function(app, user) {
 
   // Get an Incident by _id
   app.get('/api/v1/incident/:_id', user.can('view data'), function(req, res) {
+    var referer = req.headers.referer ? req.headers.referer : '';
```
pgarrison: `req.headers.referer || ''`

config/secrets.json.example
```diff
@@ -55,6 +55,7 @@
     "api": {
       "log_requests": true,
       "log_responses": true,
+      "log_user_activity": false,
```
pgarrison: We should document this (in `readme.md`?) to say that this should probably be turned on in production but won't work for the tests.

lib/api/v1/incident-controller.js
```diff
@@ -34,14 +40,20 @@ module.exports = function(app, user) {
...
     Incident
       .findById(req.params._id)
       .populate({ path: 'creator', select: 'username' })
       .populate({ path: 'assignedTo', select: 'username' })
       .exec(function(err, incident) {
         if (err) res.send(err.status, err.message);
         else if (!incident) res.send(404);
-        else res.send(200, incident);
+        else {
+          if (!(referer.split('/')[referer.split('/').length - 1] === 'reports')) {
```
pgarrison: Would it be better to check `if (referer.split('/')[referer.split('/').length - 3] === '_id')`? Not exactly sure if that works.

lib/api/v1/source-controller.js
```diff
@@ -1,8 +1,11 @@
 // Handles CRUD requests for sources.
+'use strict';
 
 var express = require('express');
 var Source = require('../../../models/source');
 var _ = require('underscore');
+var writelog = require('../../writeLog');
+
```
pgarrison: Extra line here

lib/api/v1/source-controller.js
```diff
@@ -36,6 +42,7 @@ module.exports = function(app, user) {
   // Get a Source by _id
   app.get('/api/v1/source/:_id', user.can('view data'), function(req, res) {
     Source.findByIdWithLatestEvents(req.params._id, function(err, source) {
+      // console.log(req);
```
pgarrison: !

lib/logger.js
```diff
@@ -1,8 +1,9 @@
 // Logger for the child processes
 
 var childProcess = require('./child-process');
+var userLogger = require('./userLogger');
 
-var logger = function() {};
+var logger = new Function;
```
pgarrison: I think this should just be `{}`

lib/logger.js
```diff
@@ -48,4 +48,12 @@ logger.error = function(error, metadata) {
   else childProcess.sendToParent('log', { level: 'error', message: message, metadata: metadata });
 };
 
+// Activity logger for users
+logger.logActivity = function(logData) {
+  var self = this;
+  userLogger.writeToCollection(logData, function(err) {
+    if (err) self.error(err);
```
pgarrison: And then this can just be `logger.error`

pgarrison: Actually, I think there's no need to involve `logger.js` at all. Just call `userLogger.writeToCollection` from `writeLog.js`, and if it errors you've already got `logger` in there.

lib/writeLog.js
```diff
@@ -0,0 +1,72 @@
...
+
+var config = require('../config/secrets');
+var logger = require('./logger');
+
+var logFlag = config.get().logger.api.log_user_activity;
+
+var writer = {};
+
+writer.writeIncident = function(req, incident, action) {
+  if (logFlag === false) return {};
```
pgarrison: What about `if (!logFlag) return;`? It doesn't look like the return value of `writeIncident` or any of its friends is ever used.

lib/writeLog.js
```diff
@@ -0,0 +1,72 @@
...
+      incidentTitle: incident.title,
+      incidentID: incident._id,
+      referer: req.headers ? req.headers.referer : ''
+    }
+  };
+  logger.logActivity(logDetails);
+};
+
+writer.writeSource = function(req, source, action) {
+  if (!logFlag) return {};
```
pgarrison: Similarly, `return;`?

lib/writeLog.js
```diff
@@ -0,0 +1,72 @@
...
+
+var logFlag = config.get().logger.api.log_user_activity;
+
+var writer = {};
+
+writer.writeIncident = function(req, incident, action) {
+  if (logFlag === false) return {};
+  var logDetails = {
+    userID: req.user._id,
+    username: req.user.username,
```
pgarrison: I don't see `username` in the schema. Should it be there?

lib/writeLog.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+var config = require('../config/secrets');
+var logger = require('./logger');
+
+var logFlag = config.get().logger.api.log_user_activity;
```
pgarrison: What if we just log default values for e.g. `userID`. Not sure if that's better or worse. What do you all think? @andresfib

pgarrison: Oh and if it works with the tests then we can write simple tests for it!

lib/writeLog.js
karthiksbhat: That's now. Just before the latest commit, an object used to be passed back to the controller, which would then call logActivity on it. :P

lib/writeLog.js
karthiksbhat: Oh, it should.

config/secrets.json.example
karthiksbhat: Added.

PR #223, opened 2016-11-17
5304: Looking for errors from FBgraph and downgrading them to warmings

andresfib:
No easy way to test it though

### DIFF
--- a/lib/fetching/content-services/facebook-content-service.js
+++ b/lib/fetching/content-services/facebook-content-service.js
@@ -45,11 +45,15 @@ var config = require('../../../config/secrets');
 var ContentService = require('../content-service');
 var graph = require('fbgraph');
 var util = require('util');
-var _ = require('underscore');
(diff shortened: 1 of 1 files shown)


lib/fetching/content-services/facebook-content-service.js
```diff
@@ -92,7 +96,13 @@ FacebookContentService.prototype._doFetch = function(options, callback) {
...
 
   var isTransient = function(error) {
-    return error.code && _.contains(transientErrorCodes, error.code);
+    var transient;
+    if (error.code) {
+      transient = _.includes(transientErrorCodes, error.code);
+    } else {
+      transient = _.includes(transientErrorMessages, error.message);
+    }
+    return transient;
```
pgarrison: So if there is an `error.code` but it's not transient, and the error message is one of the transient ones, we count that as not transient. I think that's fair; we can reconsider when/if those happen.

PR #224, opened 2016-11-18
5398 tag reports

andresfib:
Now we can display reports with tags! 
It includes automatic tagging of retweets and original tweets.


andresfib:
This is now in PR 226

PR #225, opened 2016-11-18
5358 multiple twitter sources

pgarrison:
Now you can create multiple Twitter sources, which will all share a single connection to the streaming API. 

If a tweet is received that matches multiple sources, all of those appear in its sources field, which is now an array (for all reports). One tweet will not become multiple reports. In the case that a report has multiple sources which have different pictures, the picture for the media column is that of the first source that still exists (i.e. if one of the multiple twitter sources is deleted, but not all, the deleted source will appear in the source field but the picture will continue to be the normal twitter picture).

Note that, as before, Aggie still receives tweets that match the keywords in fields other than the `text` field that Aggie displays (e.g. `retweeted_status.extended_tweet.full_text`). This has been slightly confusing in the past and it will continue to be confusing. So, if Aggie claims that a tweet comes from a particular twitter source despite not looking like it should, it probably matches the keywords from that source in somewhere we don't display. Looking at the original tweet on Twitter can give some insight here.

Part of this code has been pulled out into a separate module [tweet-matches](https://github.com/pgarrison/tweet-matches).


pgarrison:
We do some extra processing on each tweet now. I don't know how expensive that is, but it can run just fine at the streaming rate limit on my desktop.

PR #226, opened 2016-11-21
5267 tag sources

andresfib:
This pull requests includes also 5398 and has been rebased on develop.

### DIFF
--- a/public/angular/templates/sources/index.html
+++ b/public/angular/templates/sources/index.html
@@ -14,6 +14,7 @@
       <th class="compact" translate>Name</th>
       <th class="compact" ng-show="currentUser.can('view other users')" translate>Creator</th>
       <th class="expand" translate>Target</th>
+      <th class="expand" translate>Tags</th>
       <th class="compact" ng-show="currentUser.can('edit data')" translate>New Warnings</th>
       <th class="compact" ng-show="currentUser.can('edit data')" translate>Enabled</th>
     </tr>
@@ -32,6 +33,7 @@
         <a href="profile/{{s.user.username}}">{{ s.user.username }}</a>
       </td>
       <td class="expand content">{{ target(s) }}</td>
+      <td class="expand content">{{ tagsToString(s.tags) }}</td>
       <td class="compact content zero-errors" ng-show="currentUser.can('edit data')"
           ng-class="{ 'multiple-errors': s.unreadErrorCount > 0 }">{{ s.unreadErrorCount }}</td>
       <td class="compact status" ng-show="currentUser.can('edit data')">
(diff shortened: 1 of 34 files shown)


public/angular/templates/sources/show.html
```diff
@@ -31,6 +31,10 @@
       </td>
     </tr>
     <tr>
+      <td class="detail-key" translate>Tags</td>
+      <td class="detail-value">{{ tagsToString(source.tags) }}</td>
+    </tr>
```
pgarrison: Trailing whitespace here

shared/tags.js
```diff
@@ -0,0 +1,21 @@
+// Library for handling tags
+'use strict';
+var _ = require('lodash');
+
+// Tags are string comma separated in the front end
+var toArray = function(tags) {
+  if (typeof tags === 'string') {
```
pgarrison: So if it's not a string, we just return `undefined`? Won't that cause problems harder to debug than `object doesn't have method 'trim'`?

test/backend/models.query.report-query.test.js
```diff
@@ -23,6 +51,10 @@ describe('Query attributes', function() {
     var similar = ReportQuery.compare(otherQuery, query);
     expect(similar).to.be.true;
   });
+
```
pgarrison: Trailing whitespace here

public/angular/templates/sources/modal.html
```diff
@@ -74,6 +74,14 @@ <h4 class="modal-title">
               </div>
             </div>
           </div>
+          <div class="form-group" ng-show="source.media">
+            <label for="tags" class="col-lg-2 control-label" translate>Tags</label>
+            <div class="col-lg-8">
+              <input ng-model="source.tags" ng-list class="form-control" id="tags" name="tags" placeholder="{{ 'Enter tags' | translate }}">
+              </input>
+              <div class="help-block" translate="sources_modal.tags"></div>
```
pgarrison: This string needs to be added to `translations/locale-en.json`

public/angular/templates/sources/index.html
```diff
@@ -14,6 +14,7 @@
       <th class="compact" translate>Name</th>
       <th class="compact" ng-show="currentUser.can('view other users')" translate>Creator</th>
       <th class="expand" translate>Target</th>
+      <th class="expand" translate>Tags</th>
```
pgarrison: This column is really skinny in my browser. Should we change that?

shared/tags.js
andresfib: Good point

public/angular/templates/sources/modal.html
pgarrison: Fixed

public/angular/templates/sources/index.html
pgarrison: Documented as issue 5446

PR #236, opened 2016-11-28
5470 Report, incident tag filters ignore case

pgarrison:
For mongo's `$all` queries, we pass regular expressions to do a
case insensitive match on tags.

### DIFF
--- a/models/to-regexp.js
+++ b/models/to-regexp.js
@@ -1,0 +1,22 @@
+/*
+ * Very basic shared function between some models
+ */
+'use strict';
+
+function toRegexp(caseInsensitive, str) {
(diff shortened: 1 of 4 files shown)


models/to-regexp.js
```diff
@@ -0,0 +1,22 @@
...
+  var s = '^' + str + '$';
+  if (caseInsensitive) {
+    return new RegExp(s, 'i');
+  }
+  return new RegExp(s);
+}
+module.exports = toRegexp;
+
+var i = toRegexp.bind({}, true);
+module.exports.i = i;
```
andresfib: Do we need to export `i`?

andresfib (approved):

PR #239, opened 2016-11-29
Removed unique property for indexing user logs

karthiksbhat:
There are some user actions that trigger more than one user log action at the same time. For example, adding a report to an incident (involving creating an incident) first creates the incident, and immediately adds the report to an incident. With the unique property, there were errors as these actions used to occur within the same millisecond.

### DIFF
--- a/lib/userLogger.js
+++ b/lib/userLogger.js
@@ -13,7 +13,7 @@ var logSchema = new mongoose.Schema({
   actionRef: mongoose.Schema.Types.Mixed
 });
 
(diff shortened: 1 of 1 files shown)


andresfib:
Why did we/I come up with the idea of that unique key for the collection?

karthiksbhat:
We didn't, really. I added it thinking it wouldn't cause a problem, because there wouldn't be instances of users doing separate actions within a millisecond. What I hadn't accounted for was users' actions triggering multiple logging calls within the millisecond (ex. the one in the description of the PR).

PR #240, opened 2016-11-30
Adding additional fields to make userlogs more useful

karthiksbhat:
- Media of source/report touched added to user logs
- Storing more information when an incident is created/edited
- The filter in place for a user
- Grab batch

This PR can be kept open until the last moment, and all addition fields can be added to this branch, and later merged.

### DIFF
--- a/lib/writeLog.js
+++ b/lib/writeLog.js
@@ -20,6 +20,18 @@ writer.writeIncident = function(req, incident, action) {
       referer: req.headers ? req.headers.referer : ''
     }
   };
+  if (action === 'editIncident' || action === 'createIncident') {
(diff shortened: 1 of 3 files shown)


lib/writeLog.js
```diff
@@ -71,6 +85,48 @@ writer.writeReport = function(req, report, action) {
...
+  var logDetails = {
+    userID: req.user._id,
+    username: req.user.username,
+    action: action,
+    actionRef: {}
+  };
+  if (reports) {
+    var reportContents = [];
+    for (var report in reports) {
+      if (!reports.hasOwnProperty(report)) {
```
pgarrison: If `reports` is an array, we should just use `reports.forEach`

PR #241, opened 2016-11-30
WhatsApp Source can now be added

karthiksbhat:
We now have a jugaad (https://en.wikipedia.org/wiki/Jugaad) way to get reports from WhatsApp!
WhatsApp  content service is a subscribe-type content service, and can be scaled up to use more than one WhatsApp phone number (like any other subscribe-type cs).

Prerequisites are using Firefox browser having GNotifier plugin installed (https://addons.mozilla.org/en-US/firefox/addon/gnotifier/) and using a Linux or Windows machine. WhatsApp Web must be open, and logged into on this browser. Ensure that Desktop Notifications are enabled on WhatsApp Web. Also, ensure to navigate away from the tab/window in which WhatsApp Web is open.

There must be an established convention for the `keyword` entered while creating the source, and the identifier sent by the GNotifier plugin. A suggested convention would be having the *Phone Number* of the receiver, without any '+' or spaces.

In the GNotifier preferences, make sure to select `Custom Command` as the notification engine. Below, in the Custom command input box, enter:
```curl --data-urlencode "receiver=<phone number here>" --data-urlencode "from=%title" --data-urlencode "text=%text" http://<IP address|domain name>:2222/whatsapp```

This creates a listener for Firefox notifications when a user is away from the tab creating the notification, and sends a request to Aggie. Aggie will parse this request and generate a report.


pgarrison:
I think this shouldn't stay in develop, given its jugaadi nature. Do we want to keep supporting this after this election?

pgarrison:
Tested with `iceweasel 45.5.0esr-1~deb8u1` on `Debian 3.16.36-1+deb8u2` with `GNotifier 1.9.8`.

One step missing from the documentation here is that you have to turn desktop notification on :)

karthiksbhat:
@pgarrison Afaik, there's no WhatsApp API to do this easily and cleanly. So if we intend to have WhatsApp sources as a permanent feature of Aggie (I think it's cool), its implementation will be some hack like this.

PR #247, opened 2016-12-03
Monitors can create/edit/delete trends

karthiksbhat:
There was no permissions allotted for 'manage trends', and therefore, monitors could not create trends. This has been fixed.

### DIFF
--- a/shared/user.js
+++ b/shared/user.js
@@ -11,7 +11,7 @@
   User.PASSWORD_MIN_LENGTH = 8;
 
   User.permissions = {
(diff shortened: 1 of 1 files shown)


pgarrison:
Sorry, we just had a meeting here and decided monitors shouldn't be allowed to touch trends. :confused:

karthiksbhat:
Shall I push a commit with permissions for `manage trends` anyway (Admin only, this time)? It's ugly to make it use the default admin permission.

PR #248, opened 2016-12-03
Twitter blacklist

pgarrison:
Ignore some twitter terms.

Kingston is coming up with a list to use.


karthiksbhat (approved):
This blacklists the contents of the tweet, but not the retweeted tweet. So, tweets containing the blacklisted words still come through when they are retweeted.
Apart from that, works fine on my local system.

PR #251, opened 2016-12-05
Added comment/post tag to Facebook reports

karthiksbhat:
Since Twitter has an RT and NO_RT tag added, I feel it's required for Facebook as well, given that sometimes comments might not be as useful as posts to get relevant information.
So, I've followed Andrés' model on adding the twitter RT|NO_RT tag, and done the same for Facebook.

### DIFF
--- a/lib/fetching/content-services/facebook-content-service.js
+++ b/lib/fetching/content-services/facebook-content-service.js
@@ -150,7 +150,7 @@ FacebookContentService.prototype._handleResults = function(data) {
     reportData.push(self._parse(post));
     if (post.comments && post.comments.data && post.comments.data.length > 0) {
       post.comments.data.forEach(function(comment) {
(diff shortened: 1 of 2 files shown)


lib/fetching/content-services/facebook-content-service.js
```diff
@@ -170,7 +172,8 @@ FacebookContentService.prototype._parse = function(data) {
     likeCount: data.likes ? data.likes.summary.total_count : 0,
     reactionCount: data.reactions ? data.reactions.summary.total_count : 0,
     place: data.place ? data.place : '',
-    link: data.link ? data.link : ''
+    link: data.link ? data.link : '',
+    isComment: data.isComment ? true : false
```
pgarrison: `Boolean(data.isComment)`

PR #270, opened 2017-11-08
5543 incident map widget

andresfib:
This pull request implements a map widget for displaying public incident. For that

* Created a public endpoint for non-authorized user access to public incidents. While in the map public incidents could be rendered without the public API call, I have left the public end points as an example in case other widgets or orgs want to consume aggie's public incidents.
* Incident modal now has as expanded dialog to mark incident as public and add a public description.
image

* Incidents can be filtered by public or private.
* Settings for the public widget is set up by defining the center and the zoom level of the map.
* Documentation has been created. Maybe website should also promote the new widget when releasing the new version.

In a side change, index.html loads the gPlaces token in the same way it has been implemented for the widget map.


andresfib:
I don't think that abstracting duplication pointed by codeclimate is necessary or worth the effort.

hooverlunch:
@andresfib Exciting stuff. Could you provide some screenshots of this in operation? I haven't used my development environment in some time and would rather not fight with it.

andresfib:
Thanks @hooverlunch 
You can see the modified incident modal above.

The settings page now has a new section for widgets, only incident map at the moment.
Settings page

The settings for the map are only the center of the map and the zoom level. Everything is documented. 
Settings for incident map

 
The map is a simple HMTL which we hope that users will put in an iframe inside their webpages. When incidents are too close to each other their marker changes and clicking it will spread them. 
Incidents overlapping

Incidents spread

The incidents change color according to their veracity.
incident_map_key

We may consider bringing these map changes (color, and overlapping behaviour) to the map within aggie as well.

Travis fails because of this https://github.com/nfroidure/bufferstreams/issues/6 . I'll check if pinning a previous version of the library help.

hooverlunch:
I haven't forgotten about this, just trying to find time. Seems CI is passing now, that's good. Are all the code changes in?

andresfib:
All the changes relative to the incident map widget are here. You may have a look to other changes in v2.6.0 (#274). However, biggish commits have already been reviewed by @pgarrison and @karthiksbhat. 
I have an unpushed branch that adds the Spanish translation, which required small GUI fixes.

smoyte (changes requested):
Nice work Andres! You've come along way! This will be a very nice addition to the system!

Bunch of suggestions/requests inline, plus here are some comments on the UX:

* Can the settings section say 'Widgets' instead of 'Widget'
* Why are there two copies of the map key image in the source? Is its only use in the documentation?
* It's best to avoid putting text in an image like that. It makes it harder to translate and it doesn't look super good. In the docs, can you just have images for the map markers and then label them with text? So ideally this map key image would go away entirely.
* Can the marker colors match the Aggie color scheme? Maybe green = confirmed true, orange = confirmed false, blue = unconfirmed?

Thanks again!

hooverlunch:
PS Sorry for the big red X. I hate that feature of Github!

andresfib:
Thanks for your careful review and for helping me improve. I have followed most of your advices but:
* In Aggie, FlashService has been renamed to flash pretty much every time. 
* That's it :)

Modal in service and not in controller was how I managed to do it long time ago. I just carried it forward. Now modals are in controllers, though I think they could still be arranged a bit better.
The idea of having the image with Map Key was for people deploying Aggie to easily copy it (from the docs), or reference to it (from the deployment). I am happier with the single HTML table in the docs now however.

Regarding the colors, I followed your suggestion, which looks nice. Time will tell what users like...

Tests are passing in my machine, but Travis MongoDB version is too new. I will see to bring our MongoDB and its driver up to it.

Thanks again

PR #271, opened 2017-11-09
5584: Locking a batch is now atomic using read/write lock

andresfib:
MongoDB does not have support for atomic updates of batches. While we move to
some other DB, a locking mechanism is put in place to make sure that reports are
only checked out by one user at a time.

### DIFF
--- a/models/batch.js
+++ b/models/batch.js
@@ -3,10 +3,12 @@
 var Report = require('./report');
 var async = require('async');
 var _ = require('lodash');
+var ReadWriteLock = require('rwlock');
 var ReportQuery = require('./query/report-query');
 
 var ITEMS_PER_BATCH = 10; // 10 items per batch
 var BATCH_TIMEOUT = 5 * 60 * 1000; // 5 minutes
+var lock = new ReadWriteLock();
 
 function Batch() { /* empty constructor */ }
 
@@ -48,17 +50,25 @@ Batch.prototype.lock = function(userId, query, callback) {
     read: false
   });
 
-  Report
-    .find(filter)
-    .sort({ storedAt: -1 })
-    .limit(ITEMS_PER_BATCH)
-    .exec(function(err, reports) {
-      if (err) return callback(err);
-      var ids = _.map(reports, '_id');
-      var update = { checkedOutBy: userId, checkedOutAt: new Date() };
-      Report.update({ _id: { $in: ids } }, update, { multi: true }, callback);
-    }
-  );
+  lock.writeLock(function(release) {
+    Report
+      .find(filter)
+      .sort({ storedAt: -1 })
+      .limit(ITEMS_PER_BATCH)
+      .exec(function(err, reports) {
+        if (err) {
+          release();
+          return callback(err);
+        }
+        var ids = _.map(reports, '_id');
+        var update = { checkedOutBy: userId, checkedOutAt: new Date() };
+        Report.update({ _id: { $in: ids } }, update, { multi: true }, function() {
+          release();
+          callback();
+        });
+
+      });
+  });
 },
 
 // load a batch for user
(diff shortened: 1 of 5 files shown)


test/backend/models.batch.test.js
```diff
@@ -109,6 +135,23 @@ describe('Report', function() {
     });
   });
 
+  it('should not give same reports to different users', function(done) {
+    var batches = [];
+    createReports(function() {
+      async.filter([user._id, user2._id], function(user, cb) {
```
pgarrison: `filter` is not what you want here... either use `each` like this, or `map` and move batches out of the parent scope.

test/backend/models.batch.test.js
```diff
@@ -109,6 +135,23 @@ describe('Report', function() {
     });
   });
 
+  it('should not give same reports to different users', function(done) {
```
pgarrison: Does this test fail for the old code? I'm suspicious because it might only be a problem under high load.

models/batch.js
```diff
@@ -48,17 +50,25 @@ Batch.prototype.lock = function(userId, query, callback) {
...
-    .sort({ storedAt: -1 })
-    .limit(ITEMS_PER_BATCH)
-    .exec(function(err, reports) {
-      if (err) return callback(err);
-      var ids = _.map(reports, '_id');
-      var update = { checkedOutBy: userId, checkedOutAt: new Date() };
-      Report.update({ _id: { $in: ids } }, update, { multi: true }, callback);
-    }
-  );
+  lock.writeLock(function(release) {
```
pgarrison: Does `findAndModify` in mongo not work for this? I would expect there would be a way to do this without resorting to external locks. It's generally going to be cleaner to keep it within mongo too, since, for example, `rwlock` will probably only be able to lock within one process.

andresfib: I came to this bug happily thinking that what you said was true. Alas, ```findAndModify``` only works with single documents. ```update``` does work with several documents but does not ```sort``` nor ```limit```. The ```batch.lock``` and ```batch.load```  is a mechanism around those limitations.

There are other locking mechanism that work across processes, e.g., https://github.com/xhawk18/key_mutex , however our current ```cluster_app``` is not working.

test/backend/models.batch.test.js
andresfib: This test fails for the old code. It is an extreme case in where the ```Report.find``` in ```batch.lock``` actually returns the same ```ids``` for both users. The test fails as ```batch[0].length``` is 0: ```batch.load``` does not find any reports checked out by ```user```.  Only user2 gets the batch as the ```Report.update``` of ```user2``` overwrites the results of ```Report.update``` of user. 

In the high load scenario, where there are new reports coming in, there won't be such an extreme overlapping. Also, it can be that the ```batch.load``` of ```user``` can be executed before the ```Report.update``` of ```user2```, and, thus, the overwrite does not stop ```user``` from getting a batch, even if it is not checked out by ```user``` in the database anymore.

models/batch.js
pgarrison: Darn, okay. Would `$isolated` work here?

test/backend/models.batch.test.js
```diff
@@ -109,6 +135,22 @@ describe('Report', function() {
...
+    createReports(function() {
+      async.map([user._id, user2._id], function(user, cb) {
+        batch.checkout(user, {}, function(err, reports) {
```
pgarrison: `batch.checkout(user, {}, cb)`

test/backend/models.batch.test.js
pgarrison: :+1:

models/batch.js
andresfib: I used it and did not solve anything by itself. May help in other extreme cases, however, I removed it as I was not sure about the performance penalty it will add in MongoDB.

PR #280, opened 2020-02-11
added check for burmese encoding

harshilshah4251:
check for encoding


TheDanieLNK (commented):
Hey Harshil, this looks great.
I have removed all hard-coded values from the controller. I have added them to the config file.
Please pull latest changes, change the probability 0.9 to config.get().zawygiProb and resend pull request. Don't bother about the config file, I'll set it on the server.

PR #281, opened 2020-02-19
Harshil dev

harshilshah4251:
Added encoding code to twitter, whatsapp, rss. Need more testing though.


TheDanieLNK:
Hey Harshil, please confirm that this new update works locally. Does your Aggie instance pull correctly from Twitter and CrowdTangle?

PR #299, opened 2020-05-16
Update deps reynold2

mkyaw6:
Updated Gulp File and its relevant dependencies. Please review.


gulpfile.js
```diff
@@ -122,4 +122,4 @@ gulp.task('debugTranslations', function() {
   return merge(stream1, stream2);
 });
 
-gulp.task('build', ['debugTranslations', 'angular']);
+gulp.task('build', gulp.parallel(['debugTranslations', 'angular']));
```
karpawich: I looked through the Gulp 4 changelog and saw that the biggest change was the introduction of the `series` and `parallel` functions.

Are you sure that these tasks are all supposed to run in parallel? I'm having trouble trying to find the default behavior (series vs. parallel) that Gulp 3 had.

karpawich: (Asking because I couldn't find the answer regarding default behavior myself)

gulpfile.js
```diff
@@ -46,7 +47,11 @@ gulp.task('lint', function() {
...
 pipes.buildAngular = function() {
-  return gulp.src('public/angular/js/app.js')
+  var b = browserify({
+    entries: 'public/angular/js/app.js'
+  }).bundle();
+  return b
+    .pipe(source('public/angular/js/app.js'))
     .pipe(plumber())
     .pipe(sourcemaps.init())
     .pipe(browserify())
```
karpawich: I went down the whole `gulp` + `browserify` rabbit hole that you did, and saw the `vinyl-source-stream` [example code](https://github.com/hughsk/vinyl-source-stream) for linking browserify with gulp which matches most of what you have here. Awesome!

One comment: Do we need to call `browserify` at the end of this function chain anymore? I'm not even sure we could, since you replaced the `gulp-browserify` plugin with the actual `browserify` package.

I am probably missing something. Let me know your thoughts!

README.md
```diff
@@ -109,7 +109,8 @@ The following need to be installed.
     - If you do not have the certificate you can create a new self-signed certificate with the following command:
   `openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365`
     - This will allow you to start the server but it will generate unsafe warnings in the browser. You will need a real trusted certificate for production use.
-    - Adding the `-nodes` flag will generate an unencrypted private key, allowing you to run tests without going through a password prompt
+    - Create a passphrase.txt file under config and enter in the PEM passphrase
```
karpawich, line 112: We should try to avoid the storage of our private key decryption passphrase in plaintext on the filesystem. It does not matter for development, but our production builds need to be secure! If someone were to be granted access to the contents of the passphrase file, they could decrypt all HTTPS communications.

Let's explore another solution. Off the top of my head, we could get the PEM passphrase prompt to work natively, or we could just build in our own prompt.

gulpfile.js
mkyaw6: We don't need to call browserify anymore. gulp-browserify is no longer supported and I used a 'recipe' that was provided in the docs to have the same behavior.

gulpfile.js
mkyaw6: I am also not 100% sure what the default behavior was. We would need to verify it with the rest of the team if anyone knows whether it should be series or parallel

README.md
mkyaw6: I agree. This was mainly a stop-gap solution for the sake of time.

mkyaw6: The PEM passphrase was not being prompted and I spent a lot of time trying to figure out how to make that happen, so I implemented a temporary solution.

gulpfile.js
karpawich: I found [this guide](https://codeburst.io/switching-to-gulp-4-0-271ae63530c0) to be helpful in describing the original behavior. I'll use it to make sure we preserve Gulp behavior!

karpawich (approved):

Issue #316, opened 2020-05-19
Update Mongoose from 3.8.39 to latest


karpawich:
Express 3.8.39 and corresponding plugin packages introduce vulnerabilities to Aggie, including several of high severity. To resolve these vulnerabilities, we need to update Mongoose and corresponding plugins to their latest versions. 

Some plugins which are now unmaintained will need to be replaced with a maintained alternative, namely `mongoose-auto-increment`. Others may also need to be replaced as we perform the upgrade.

_Some preliminary work has already been done to update Mongoose. See commits below._
- [59a08b](https://github.com/TID-Lab/aggie/commit/59a08b49badc469cfe1bac1cec8b302583029f2d)

karpawich:
Mongoose 3.8 also has documented installation problems that require workarounds:

https://github.com/Automattic/mongoose/issues/2718

PR #322, opened 2020-06-09
Upgrade Node from 0.10 to 12.6

### LINKED ISSUE #294: Update backend & frontend dependencies
## Introduction

Before major development can resume on the Aggie repo, dependencies should be updated to their latest versions so that we avoid relying on outdated, unstable, or unsupported libraries and frameworks.

- **Node:** 0.10 to 12.16.3

karpawich:
### Resolves #294


karpawich:
[Upgrade modules/dependencies (node 12 and angular 1.7 and whatever else required for that)](https://trello.com/c/cBqAeHnx/13-upgrade-modules-dependencies-node-12-and-angular-17-and-whatever-else-required-for-that)

install.js
```diff
@@ -13,19 +13,19 @@ function enableIndexing(callback) {
...
+          useNewUrlParser: true,
+          useUnifiedTopology: true,
+          useCreateIndex: true,
+        });
     }, 200);
   });
   database.mongoose.connection.once('open', function() {
-    // Enable database-level text search
-    database.mongoose.connections[0].db.admin().command({ setParameter: 1, textSearchEnabled: true }, function(err, res) {
+    Report.ensureIndexes(function(err) {
```
harshilshah4251, line 25: Seems like this function is deprecated after MongoDB 3.0.0 and just serves as an alias to create index. Maybe just change this to createIndex in case they totally remove ensureIndex in future. https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex/index.html

install.js
```diff
@@ -13,19 +13,19 @@ function enableIndexing(callback) {
...
     setTimeout(function() {
-      database.mongoose.connect(database.connectURL);
+      database.mongoose.connect(database.connectURL,
+        {
+          useNewUrlParser: true,
+          useUnifiedTopology: true,
+          useCreateIndex: true,
+        });
     }, 200);
   });
```
harshilshah4251, line 23: If this returns successfully, then print out a message "Text search has been enabled for MongoDB" as it was in the older version. Also couldn't find anything about `useNewUrlParser: true,
          useUnifiedTopology: true,
          useCreateIndex: true,` online but I assume it does the same thing as `    database.mongoose.connections[0].db.admin().command({ setParameter: 1, textSearchEnabled: true }, function(err, res) {`

lib/api.js
```diff
@@ -88,15 +101,15 @@ function logRequests(req, res, next) {
...
+    logger.debug(`>>>> [${requestId}, ${clientIP}] ${req.method} ${req.url}`);
   }
 
   if (responseLoggingEnabled) {
     var end = res.end;
     res.end = function(chunk, encoding) {
       res.end = end;
       var duration = new Date() - start;
-      logger.debug(strformat('<<<< [{0}, {1}] {2} {3} {4} {5}ms', requestId, clientIP, req.method, req.url, res.statusCode, duration));
+      logger.debug(`>>>> [${requestId}, ${clientIP}] ${req.method} ${req.url} ${req.statusCode} ${duration}ms`);
```
harshilshah4251, line 112: I understand that `consol.log` prints out to the terminal and logger prints out log messages in the files in the logs directory. Is there any sort of consensus over what should be printed using `console.log` vs what should be printed in the log files?

lib/api/reset-password.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(app, auth) {
   app = app || express();
   auth = auth || authentication(app);
 
-  app.use(express.urlencoded());
+  app.use(express.urlencoded({extended: true}));
```
harshilshah4251: What does this added option do? Maybe add a comment here?

lib/api/socket-handler.js
```diff
@@ -2,7 +2,7 @@
 // Watches for updates to sourceErrorCount
 
 var express = require('express');
-var socketIo = require('socket.io');
+var SocketIOServer = require('socket.io');
```
harshilshah4251, line 5: naming convention? SocketIOServer->socketIOServer

lib/api/socket-handler.js
```diff
@@ -98,19 +100,8 @@ SocketHandler.prototype.emitAllErrorsCount = function(emitter) {
 SocketHandler.prototype._configureSocketIO = function() {
```
harshilshah4251, line 97: add a comment here why you chose to remove lines 101-111. If its not relevant to write a comment in the file, you can just explain it here for my understanding.

lib/api/socket-handler.js
```diff
@@ -16,24 +16,26 @@ var Trend = require('../../models/trend');
...
-var emit = socketIo.Socket.prototype.emit;
-socketIo.Socket.prototype.emit = function(event, data) {
-  logger.debug(event, data);
-  emit.apply(this, arguments);
-};
-
-
+/**
+ * 
+ * TODO the Socket object prototype is not exposed in v2.3.0
```
harshilshah4251: Is this still a TODO or has it been implemented somewhere else?

lib/api/streamer.js
```diff
@@ -61,7 +61,10 @@ Streamer.prototype.query = function() {
         if (allEmpty && self.status === 'querying') self.status = 'idle';
         // If not idle, queue next query batch
         if (self.status !== 'idle') {
-          setTimeout(self.throttledQuery(), QUERY_INTERVAL);
+          var throttledQuery = self.throttledQuery()
+          if (typeof throttledQuery !== 'undefined') {
+            setTimeout(throttledQuery, QUERY_INTERVAL);
```
harshilshah4251, line 69: Was writing `setTimeout(self.throttledQuery(), QUERY_INTERVAL);` directly giving you an error?

lib/fetching/content-services/elmo-content-service.js
```diff
@@ -4,7 +4,7 @@ var ContentService = require('../content-service');
 var util = require('util');
 var config = require('../../../config/secrets');
 var request = require('request');
-var dateFormat = require('dateformat');
+var moment = require('moment');
```
harshilshah4251, line 7: design decision to use moment over dateformat? just curious.

lib/process-manager.js
```diff
@@ -17,7 +17,7 @@ var ProcessManager = function() {
 // @modulePath Path of the module relative to the root directory
 ProcessManager.prototype.fork = function(relativePath) {
   var self = this;
-  var absolutePath = require.resolve(path.join(process.env.PWD, relativePath));
+  var absolutePath = require.resolve(path.join(__dirname, '..', relativePath));
```
harshilshah4251, line 20: AFAIK, `process.env.PWD` get the PWD from docker-compose file if its present. Where does it fetch the value for `__dirname` ? Is it like a global variable that always defined?

models/incident.js
```diff
@@ -74,10 +69,10 @@ schema.post('remove', function() {
 
 });
 
+schema.plugin(AutoIncrement, { inc_field: 'idnum' });
```
harshilshah4251, line 72: mongoose sequence automatically creates this and still requires us to AutoIncrement?

models/source.js
```diff
@@ -6,21 +6,23 @@
...
-  validator: 'isURL',
-  passIfEmpty: true
-});
+var lengthValidator = function(str) {
+  return validator.isLength(str, {min: 0, max: 20})
+}
+
+var urlValidator = function(url) {
+  return (
+    (typeof url === 'undefined' || url === null)
```
harshilshah4251: People seem to recommend using `myVariable === undefined`. Up to you if you want to change.https://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript

harshilshah4251: Also I might be misunderstanding the functionality of this urlvalidator but seems like something is off. If a url is an empty string, then urlValidator would be assigned boolean value True. Shouldn't it assign boolean value False because the url is empty and hence invalid?

lib/ses-transport.js
```diff
@@ -0,0 +1,54 @@
+// custom Winston transport for AWS SES (Simple Email Service) using the AWS JavaScript SDK
+// based a now-outdated outdated package https://github.com/jpgarcia/winston-amazon-ses
+
+const Transport = require('winston-transport');
```
harshilshah4251, line 4: How is this new file being used?

harshilshah4251:
Also, why does the Travis CI still fail after adding the mocha.opts file? I thought that fixed the issue once for at least one commit earlier.

harshilshah4251:
Also getting this while running Aggie `(node:67) DeprecationWarning: collection.count is deprecated, and will be removed in a future version. Use Collection.countDocuments or Collection.estimatedDocumentCount instead`. I think its not a problem with the code in Aggie but i think it refers to outdated code in some package. https://stackoverflow.com/questions/51146152/mongodb-nodejs-driver-deprecationwarning-collection-count-is-deprecated. Dont think we should worry about that now but it might come up in future.

install.js
karpawich: Your assumption is correct - since MongoDB 2.6, text search is enabled by default internally.

install.js
karpawich: Since Aggie still works with these deprecated functions, we made a separate Issue & Trello task to update them later. See #316

lib/api.js
karpawich: No, although that is a great question.

In production, Aggie will also log any errors to a Slack channel and an email address that we point it to. I would guess that the idea behind logging everything to files is so that the logs persist on disk for someone to view later. During production, it will be much harder to see `console.log` output since we are not running it from our own terminal.

lib/api/reset-password.js
karpawich: That option does nothing. It is for a later version of the `urlencoded` middleware that we decided not to upgrade to. See https://github.com/expressjs/body-parser#extended

I've removed all instances of adding `extended: true` from the branch :+1:

lib/api/socket-handler.js
karpawich: This is still a TODO! Thank you for catching it.

They manipulated the `socketio.Socket` prototype to log events, [but that is now considered bad practice](https://codeburst.io/what-is-prototype-pollution-49482fc4b638) because it can introduce a vulnerability.

I will need to look into how to achieve the same according to modern conventions

lib/api/socket-handler.js
karpawich: Sure! I'll explain my reasons here:

1. Each `io.enable` call is now handled internally by SocketIO
2. **I will need to re-investigate `io.set('log-level', 1)`** - I'm not sure I ever replaced it. TLDR: [SocketIO uses a completely different system for logging now](https://socket.io/docs/migrating-from-0-9/#Log-differences)
3. The `io.set` call at line 109 is configured internally by SocketIO now, and only two transports are available. By default, 2/2 of those transports are used, so there is no need to manually enable them all myself like the old maintainers did here.

Finally, `io.configure` is also gone, since SocketIO replaced `io.enable` and `io.set` calls with alternative mechanisms, so I removed the entire section altogether.

lib/api/streamer.js
karpawich: Sometimes `throttledQuery` would be undefined. In that case, `setTimeout` would throw an error. I introduced a check so that `throttledQuery` would only be called when it is actually defined.

lib/fetching/content-services/elmo-content-service.js
karpawich: Before, the repo used both `dateformat` and `moment` separately. Because `moment` can do everything we used `dateformat` for, I just got rid of `dateformat`.

lib/process-manager.js
karpawich: Yes! `__dirname` [is the modern standard for path resolution in Node](https://nodejs.org/docs/latest/api/modules.html#modules_dirname).

lib/ses-transport.js
karpawich: I wrote a custom Winston transport for AWS SES to replace the **7 year-old** `winston-amazon-ses` package.

[Given how simple the old one was](https://github.com/jpgarcia/winston-amazon-ses/blob/master/lib/winston-amazon-ses.js), I decided that the best approach would be to rewrite myself using the latest Winston and AWS Node SDK.This file is the new SES transport.

models/incident.js
karpawich: This line _is_ where `mongoose-sequence` "automatically" creates the `idnum` field.

models/source.js
karpawich: [Aggie allowed empty strings before](https://github.com/TID-Lab/aggie/blob/06893171d1b1a7fbc0cb57d887c82c84c8355dc9/models/source.js#L22), so I just replicated the functionality. Changing that functionality so that empty URL's are considered invalid would be a separate Issue altogether.

karpawich:
@harshilshah4251 that is another MongoDB deprecation notice. See #316

README.md
```diff
@@ -205,7 +203,7 @@ Aggie uses Google Places for guessing locations in the application. To make it w
 ### Emails
   1. `fromEmail` is the email address from which system emails come. Also used for the default admin user.
   1. `email.from` is the address from which application emails will come
-  1. `email.transport` is the set of parameters that will be passed to [NodeMailer 2.1.0](http://www.nodemailer.com). Valid transport method values are: 'SES', 'sendgrid' and 'SMTP'.
+  1. `email.transport` is the set of parameters that will be passed to [NodeMailer 6.4.6](http://www.nodemailer.com). Valid transport method values are: 'SES', 'sendgrid' and 'SMTP'.
```
cooperka: May as well remove the version string here 🤷‍♂️

config/secrets.js
```diff
@@ -55,7 +53,7 @@ module.exports.get = function(options) {
 // update fetching flag
 module.exports.updateFetching = function(flag, cb) {
   cb = cb || function() {};
-  nconf.set('fetching', S(flag).toBoolean());
+  nconf.set('fetching', !!flag);
```
cooperka, line 56: 😂 thank you for all this cleanup

lib/api.js
```diff
@@ -37,23 +35,38 @@ try {
...
+    // Prompts for passphrase
+    var passphrase = readLineSync.question("Enter PEM passphrase ", {hideEchoBack: true})
+    try {
+        server = require('https').createServer({
+          key: fs.readFileSync(keyFile),
+          cert: cert,
+          passphrase: passphrase
+        }, app);
+    } catch (error) {
+      console.log('WRONG PASSPHRASE');
```
cooperka: Consider using sentence case. Could also consider using `'TRY AGAIN LOSER'`

lib/api/streamer.js
cooperka: Noting that this is smelly (not your fault), maybe we should add a `TODO` comment here explaining that we don't know why it's sometimes undefined.

models/report.js
```diff
@@ -26,9 +25,6 @@ var schema = new Schema({
   checkedOutAt: { type: Date, index: true }
 });
 
-// Give the report schema text search capabilities
-schema.plugin(listenTo);
```
cooperka, line 30: I imagine text search capabilities happen elsewhere now, can you briefly explain?

models/source.js
```diff
@@ -6,21 +6,23 @@
...
+    || (typeof url === 'string' && url.length === 0)
+    || validator.isURL(url)
+  );
```
cooperka: Could we simply `url == null || url === '' || validator.isURL(url)`?

(Note double `== null` checks both null/undefined and has become rather conventional in JS)

cooperka: Also, will this break validator if `url` is not a string?

package.json
```diff
@@ -18,102 +18,90 @@
     "protractor": "protractor test/end-to-end/protractor.conf.js",
     "protractor-with-apis": "protractor test/end-to-end/with-apis/protractor.conf.js"
   },
+  "subdomain": "aggie",
```
cooperka, line 22: ~I can't readily find docs online, what is `subdomain`?~

Oh this was not something you added.

test/backend/lib.api.socket-handler.test.js
```diff
@@ -13,15 +13,27 @@ var express = require('express');
...
+      cert: fs.readFileSync(certFile),
+    }, app);
+  } catch {
+    // Prompts for passphrase
+    server = require('https').createServer({
+      key: fs.readFileSync(keyFile),
+      cert: fs.readFileSync(certFile),
+      passphrase: readLineSync.question("Enter PEM passphrase: ")
+    }, app);
+  }
```
cooperka: This seems like a nice chunk of duplication that could be DRYed up

test/mocha.opts
```diff
@@ -0,0 +1,3 @@
+# mocha.opts
+
+  --exit
```
cooperka, line 10: Presumably this is the "quick fix" described in [mocha docs](https://mochajs.org/#-exit), can we at least add a TODO comment?

cooperka (commented):
General pass over the code looks awesome. Just left comments about cleanup mostly. I did not verify any functionality (though it did install on my machine and it runs!)

models/report.js
karpawich: I investigated this comment thoroughly, and found no traces of how the `listen-to` custom plugin [written by an Aggie contributor](https://github.com/seelio/mongoose-listento) enables text search in the latest Mongoose.

The line below this, actually, is [_the_ way](https://docs.mongodb.com/manual/core/index-text/#overview) to support text search: you add a `text` index to a field in a MongoDB collection so that it is searchable via the [`$text`](https://docs.mongodb.com/manual/reference/operator/query/text/#op._S_text) query operator.

lib/api/streamer.js
karpawich: Yes, and yes. A TODO comment will fit the bill here so that we can return to it at a later time, if time allows 🤞

lib/api.js
karpawich: Ahahahahah 🤣 

I was not aware that this `console.log` was in the code 😅 we can definitely use proper casing (and probably not `console.log` _directly_)

config/secrets.js
karpawich: Yeah this package had strange conventions for _sure_

README.md
karpawich: Yep! Can't hurt one bit. I suppose I took the idea of preserving behavior a _little_ to seriously here

test/mocha.opts
karpawich: Yep! Thank you for all these TODO recommendations! I do like me some good ol' sustainable open-source practices 👍 🎉

test/backend/lib.api.socket-handler.test.js
karpawich: Absolutely. On it!

models/source.js
karpawich: I'm _almost_ positive that my custom validator preserves every bit of behavior from the `mongoose-validator` package.

The operative word here is _almost_. I will need to re-investigate whether my custom validator really does preserve behavior like I thought it did.

And thanks for the conventions tip! Good to know 😄

models/source.js
TheDanieLNK: Yes. Not all sources would have associated URLs, and that's fine.

docker/docker-compose.yml
```diff
@@ -1,13 +1,21 @@
...
-  environment:
-    - MONGO_HOST=db
-  ports:
-    - "443:3000"
-db:
-  image: mongo:2.6.11
-  ports:
-     - "27020:27017"
-  command: "--smallfiles --logpath=/dev/null"
+version: '3'
```
TheDanieLNK, line 1: Finally - an updated docker!

docs/content/usingaggie/usingaggie.md
```diff
@@ -319,7 +319,7 @@ To create a trend visualization:
 
 3.  Enter the *keyword(s)* you want to create the trend for. For example *NDC, NPP*.
 
-    ![Trend Keywordss](trends_keywords.png)
+    ![Trend Keywords](trends_keywords.png)
 
 4.  Choose a *Social Media Source* for the trend e.g. Twitter, RSS or Facebook.
```
TheDanieLNK, line 324: Remove RSS from this list as it is not used on Aggie anymore and a comeback is unlikely.

lib/ses-transport.js
```diff
@@ -0,0 +1,54 @@
...
+
+    aws.config.update({
+      accessKeyId: opts.accessKeyId,
+      secretAccessKey: opts.secretAccessKey,
+      region: opts.region,
+    })
+
+    var transport = {
+      SES: new aws.SES({
+        apiVersion: '2010-12-01'
```
TheDanieLNK, line 27: Not sure why this version name has 2010; same as in mailer.js

models/query.js
```diff
@@ -1,9 +1,6 @@
 // Represents a request for a subset of objects (reports, incidents, etc.; see subclasses).
 // For example, a query might be 'All reports with status = relevant and keyword foo'
 
-var database = require('../lib/database');
-var mongoose = database.mongoose;
-
```
TheDanieLNK, line 6: Looks like database and mongoose were never previously used here?

test/backend/lib.api.streamer.test.js
```diff
@@ -57,6 +57,6 @@ describe('Streamer', function() {
     });
   });
 
-  after(utils.wipeModels([Report]));
-  after(utils.expectModelsEmpty);
+    after(utils.wipeModels([Report]));
+    after(utils.expectModelsEmpty);
```
TheDanieLNK: Trivial but I think the spacing should be in the previous location to align with the 'before' logic. Thoughts?

TheDanieLNK (commented):
Made minor comments. Looks great overall!

lib/ses-transport.js
karpawich: Even according to the official AWS SDK, we should still be using the 2010 SDK:

https://docs.aws.amazon.com/ses/latest/APIReference/Welcome.html

models/query.js
karpawich: Yes.

test/backend/lib.api.streamer.test.js
karpawich: Sure. Fixed 👍

harshilshah4251:
Looks good to me! Is Travis CI returning successfully now? I cant see it here.

docker/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN mv config/secrets.json.example config/secrets.json
 RUN openssl req -x509 -newkey rsa:2048 -keyout config/key.pem -out config/cert.pem -days 365 -nodes -subj "/O=Aggie"
 
 # Uncomment when using git clone
-RUN npm install --global gulp@3 forever@0 && npm install --unsafe-perm=true
+RUN npm install --global gulp@4 forever@3 && npm install --unsafe-perm=true
```
harshilshah4251, line 15: Did you check if these versions work?

karpawich: Yes! They work on my machine 😄

PR #325, opened 2020-06-11
CrowdTangle Integration

limmabean:
This comes with the integration of the crowdtangle API into the aggie platform. However, not everything is finished. We want to add several fields into the metadata field of each report. See lib/fetching/content-services/crowd-tangle-content-service.js to see the brunt of the added code. The included json file (config/crowdtangle_list.json) is to store crowdtangle lists and their corresponding account ids. I also removed prettyjson from crowd-tangle-content-service.js and winston slack report from master-logger.js because they didn't work with a clean install of the branch. This may cause problems later.


karpawich:
I've added myself as a reviewer so that I can raise any concerns that I might have about merging this in with the Node 12 upgrades. I already see a few conflicts that we will have to address; I'll leave comments about those.

Is someone against me running `git merge update-deps` on this branch?

@limmabean @TheDanieLNK @harshilshah4251

test/backend/fixtures/ct-2.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
+    "id": 1,
+    "submitter": "Tester",
+    "created_at": "2019-11-07
```
karpawich, line 5: @TheDanieLNK Is this JSON file intentionally invalid? It appears that the file got cut off.

public/angular/translations/locale-en.json
```diff
@@ -392,11 +392,22 @@
   "Whatsapp": "WhatsApp",
   "whatsapp": "WhatsApp",
   "gplaces": "Google Places",
+  "CrowdTangle": "crowdtangle",
+  "crowdtangle" : "CrowdTangle",
+  "CrowdTangle": "CrowdTangle",
+  "Crowdtangle" : "CrowdTangle",
```
karpawich: Why are there four different capitalization's of "crowdtangle" here?

public/angular/templates/sources/modal.html
```diff
@@ -39,24 +39,35 @@ <h4 class="modal-title">
...
               <div ng-show="source.media == 'twitter'" class="help-block" translate="sources_modal.help"></div>
               <div class="has-error" ng-show="showErrors() && form.keywords.$invalid">
-                <div class="help-block" ng-show="form.keywords.$error.required">Keywords are required for twitter sources.</div>
+                <div class="help-block" ng-show="form.keywords.$error.required">Keywords are required for Twitter sources.</div>
               </div>
             </div>
           </div>
 
-          <div class="form-group" ng-show="source.media && source.media != 'twitter' && source.media != 'smsgh' && source.media != 'whatsapp'">
+          <!-- <div class="form-group" ng-show="source.media && source.media != 'twitter' && source.media != 'crowdtangle' && source.media != 'smsgh' && source.media != 'whatsapp'">
```
karpawich: Why are these blocks commented out, & can they be removed?

public/angular/js/controllers/sources/index.js
```diff
@@ -19,11 +19,12 @@ angular.module('Aggie')
     };
 
     $scope.target = function(source) {
-      return source.media == 'twitter' || source.media == 'smsgh' || source.media == 'whatsapp' ? source.keywords : source.url;
+      return source.media == 'twitter' || source.media == 'smsgh' || source.media == 'whatsapp' || source.media == 'crowdtangle' ? source.keywords : source.url;
     };
 
     $scope.sourceClass = function(source) {
-      var mediaOptions = ['twitter', 'facebook', 'rss', 'elmo', 'smsgh', 'whatsapp'];
+      var mediaOptions = ['twitter', 'crowdtangle'];
```
karpawich: @TheDanieLNK Why are these the only available `mediaOptions` for Aggie? I understand that Twitter and Crowdtangle are the two sources that the Myanmar group will use, but maybe we could move all Myanmar-specific changes to their own branch?

public/angular/js/controllers/sources/form_modal.js
```diff
@@ -30,6 +30,7 @@ angular.module('Aggie')
           flash.setNotice('source.create.success');
           $rootScope.$state.go('source', { id: response._id }, { reload: true });
         }, function(err) {
+          console.log(err);
```
karpawich: Does this `console.log` have to be there? I'm guessing that it was added for debugging purposes, but no longer is needed.

public/angular/js/controllers/reports/index.js
```diff
@@ -372,14 +375,19 @@ angular.module('Aggie')
       // Pick one of the sources that has a media type. For now, it happens that
       // if a report has multiple sources, they all have the same type, or are
       // deleted
-      for (var i = 0; i < report._sources.length; i++) {
-        var sourceId = report._sources[i];
-        var source = $scope.sourcesById[sourceId];
-        if (source && $scope.mediaOptions[source.media] !== -1) {
-          return source.media + '-source';
+
+      if (report.metadata.platform === "Facebook") {
```
karpawich, line 379: Why does this exception from the for loop have to exist for Facebook?

lib/master-logger.js
```diff
@@ -23,7 +22,6 @@ function _configureLogger(name) {
   var logger = new (winston.Logger)({
     transports: [
       new winston.transports.File(config.logger.file),
-      new winston.transports.Slack(config.logger.Slack),
```
karpawich: We can still use Slack for error logging! The Node 12 upgrades use the API (and a different package) - see the errors that got posted recently to the #aggie-deployments channel 😄 

Also see #326 for how we can making debugging errors in development easier (without having to send them to a Slack channel)

config/secrets.json.example
```diff
@@ -1,91 +0,0 @@
-{
```
karpawich: Why was this file deleted? Based on the current README, developers use this file as the baseline for their `secrets.json`

package.json
```diff
@@ -104,7 +105,9 @@
     "karma-phantomjs-launcher": "^1.0.1",
     "merge-stream": "^1.0.0",
     "mocha": "^2.3.3",
+    "myanmar-tools": "^1.1.3",
```
karpawich: Why is `myanmar-tools` listed in both the `dependencies` and `devDependencies`? If it is included in the `dependencies`, then it will install for both development _and_ production

test/backend/lib.fetching.content-service.crowd-tangle-content-service.test.js
```diff
@@ -0,0 +1,105 @@
...
+        expect(reportData.metadata.platform).to.equal('Facebook');
+        expect(reportData.metadata.type).to.equal('native_video');
+        break;
+      case 3:
+        return done(new Error('Unexpected report'));
+      }
+    });
+
+    // Give enough time for extra report to appear.
+    setTimeout(function() { if (fetched == 2) done(); }, 100);
```
karpawich: Can the `setTimeout` be substituted with an alternative like `process.nextTick` or `setImmediate` that guarantees the extra report appears before execution? I've seen many Aggie tests take advantage of those functions (among others) to do that

test/backend/lib.fetching.content-service.crowd-tangle-content-service.test.js
```diff
@@ -0,0 +1,105 @@
...
+    service.once('error', function(err) { done(err); });
+
+    service.on('report', function(reportData) {
+      expect(reportData.metadata.crowdtangleId).to.not.be.undefined;
+    });
+    setTimeout(done, 500);
+  });
+
+
+  // TODO 1
```
karpawich: Why is this TODO here? Can it be removed, or should another test be in its place?

karpawich (changes requested):
I left a number of comments. Here is a summary of my review below:

- Slack error logging with Winston works in the Node 12 upgrade branch, so we don't need to get rid of the Slack Winston transport! I used a different package that works with the latest Slack API. You can see it working in the #aggie-deployments Slack channel

- We should move all Myanmar-specific changes to their own branch. See #329 for a more detailed explanation:
  - any use of `myanmar-tools` would be moved to this branch
  - the main branch would still be able to support more than Twitter & CrowdTangle

- There may be simple frontend, end-to-end, or other backend tests that we would want to include to ensure CrowdTangle support remains stable. I also want to see if the `setTimeout` call could be replaced with something more reliable.

Looks great! Just a few suggestions 😀

harshilshah4251:
+1 on moving changes to Myanmar specific branch

lib/master-logger.js
TheDanieLNK: Great! I did see the errors logged on Slack. We will use your updated master-logger when we merge. +1 to your proposal for error logging!

package.json
TheDanieLNK: Oops, thanks for catching that! Will update.

public/angular/js/controllers/sources/form_modal.js
TheDanieLNK: You are right :) Would remove.

public/angular/js/controllers/sources/index.js
TheDanieLNK: Yes. I agree we should have a Myanmar-specific branch. I will re-enter the previous options and comment them out. No pressure though since some of these sources (SMSGh, RSS) won't be use anytime soon anyway.

public/angular/translations/locale-en.json
TheDanieLNK: Would clean that up. I was testing stuff a while ago.

test/backend/fixtures/ct-2.json
TheDanieLNK: Yes. Just a test for malformed results.

test/backend/lib.fetching.content-service.crowd-tangle-content-service.test.js
TheDanieLNK: Will remove.

public/angular/js/controllers/reports/index.js
TheDanieLNK: The exception is to handle the fact Facebook is a 'sub-source' within Crowdtangle, and to show the Facebook logo accordingly, instead of the CrowdTangle logo which is in the source.media listing.

public/angular/templates/sources/modal.html
TheDanieLNK: Removed them.

test/backend/lib.fetching.content-service.crowd-tangle-content-service.test.js
TheDanieLNK: Thanks Max. Substituting `setTimeout` with `process.nextTick`.

karpawich (approved):

lib/fetching/bots/pull-bot.js
```diff
@@ -8,7 +8,7 @@ var logger = require('../../logger');
 // options.source - The source to pull from.
 // options.contentService - The contentService to control.
 var PullBot = function(options) {
-  this.interval = options.interval || 30000;
+  this.interval = options.contentService.interval || 120000;
```
harshilshah4251, line 11: I might be missing the context but what was the reason for this change?

harshilshah4251:
Note: We should research some checkstyle/formatting tool for JS/HTML code. Its not useful to see changes such as new lines/deleted lines that happen naturally while making actual code changes. It would be nice to have a checkstyle/formatting tool so that new changes only reflect the actual code changes and not the line additions/subtractions.

lib/fetching/bots/pull-bot.js
karpawich: The original interval was 120000, so I returned it to the default and set up a way to configure the interval via a content provider so that CrowdTangle can set the interval to 30000.

PR #327, opened 2020-06-11
Instruct admins to configure Aggie email settings on email failure

### LINKED ISSUE #308: When new user is added an error message is shown and user table not automatically updated
Attempting to add users results in an error message as seen below. 
Screenshot 2020-05-19 12:46:54

The user table does not get automatically updated when a new user is added. To see newly added users, the page needs to be reloaded. This means that it seems like adding a user fails (especially considering the error message) but in reality, the user was added and can be confirmed by trying to add a user with the same name. See the error message below. 

UserIsInDatabase

karpawich:
Resolves #308

This PR adds a new error message for when a user account is created, but their password reset email fails to send. The message directs the admin to visit the Email section of the Settings page:

Screenshot (181)

This PR also makes secondary changes:
https://github.com/TID-Lab/aggie/commit/3aaf0dd0ef51359e0cebd5eae5098f32b54c9140
- Fix use of the Underscore `template` function so that data is correctly passed in [per the changes made in v1.7.0](https://underscorejs.org/#changelog)

### DIFF
--- a/lib/api/v1/user-controller.js
+++ b/lib/api/v1/user-controller.js
@@ -64,7 +64,7 @@ module.exports = function(app, user) {
       else {
         // Send password reset email
         sendEmail(user, req, function(err) {
-          if (err) res.send(err.status, err.message);
+          if (err) res.send(502, err.message); // send status code "Bad Gateway" to indicate email failure
           else res.send(200, user);
         });
       }
--- a/lib/mailer.js
+++ b/lib/mailer.js
@@ -108,10 +108,9 @@ Mailer.prototype.sendFromTemplate = function(options, callback) {
                function mailTemplate(err, template) {
                  if (err) return callback(err);
                  var subject = template.subject;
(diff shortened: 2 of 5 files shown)


karpawich:
[When new user is added an error message is shown and user table not automatically updated](https://trello.com/c/81NaXCE6/45-when-new-user-is-added-an-error-message-is-shown-and-user-table-not-automatically-updated)

mkyaw6:
The issue in [#308](https://github.com/TID-Lab/aggie/issues/308) regarding error messages shown even when a new user is successfully created has not been addressed. 
When creating a new user, an error message is still shown and page needs to be refreshed before the new user is displayed.
Screen Shot 2020-06-26 at 3 20 11 PM

karpawich:
Tests are not needed - the existing frontend, backend, and end-to-end tests for User creation do not cover error cases.

PR #335, opened 2020-06-26
Improve accessibility of the Aggie platform

### LINKED ISSUE #334: Improve accessibility of the Aggie platform
Since Aggie was written, the software development community has created a number of standards and tools for accessibility. Aggie should be compared against these standards through available tools so that its accessibility can be evaluated and improved.

karpawich:
Resolves #334


cooperka:
#387 went a long way with some of this; I'm going to close this PR since it's quite stale but leave the GH issue open in case we want to revisit later.

PR #336, opened 2020-07-10
Fix ct bugs

mkyaw6:
This PR fixes 2 bugs:
1) Fetching posts based on last report date does not work when it is the first time fetching, due to last report date being based on previously fetched reports.

2) A bug that prevents new CT sources from being created due to the frontend requiring the URL field.

### DIFF
--- a/lib/fetching/content-services/crowd-tangle-content-service.js
+++ b/lib/fetching/content-services/crowd-tangle-content-service.js
@@ -80,7 +80,9 @@ CrowdTangleContentService.prototype._httpRequest = function(params, callback) {
 CrowdTangleContentService.prototype._completeUrl = function() {
 	// add one second to last report date to start fetching from
 	// if it is the first time fetching data, initialize start date to an old date
-	var startDate = this._lastReportDate ? new Date(this._lastReportDate.getTime() + 1000).toISOString() : new Date("January 1, 1995");
+	var initialStartDate = new Date();
+	initialStartDate.setHours(initialStartDate.getHours() - 3);
+	var startDate = this._lastReportDate ? new Date(this._lastReportDate.getTime() + 1000).toISOString() : initialStartDate;
 	return url.format({
 		protocol: 'https',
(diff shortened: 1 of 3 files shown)


lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -79,7 +79,8 @@ CrowdTangleContentService.prototype._httpRequest = function(params, callback) {
 
 CrowdTangleContentService.prototype._completeUrl = function() {
 	// add one second to last report date to start fetching from
-	var startDate = new Date(this._lastReportDate.getTime() + 1000).toISOString();
+	// if it is the first time fetching data, initialize start date to an old date
+	var startDate = this._lastReportDate ? new Date(this._lastReportDate.getTime() + 1000).toISOString() : new Date("January 1, 1995");
```
TheDanieLNK: I see what you did here; looks great. However, I think we might need the logic to be different. If lastReportDate is null (meaning first fetch), we should not go very far back in time (as you have here), else we may never get to the most recent content. I'd rather propose that we set lastReportDate to be => now - someShortIntervalInHours (where 1 <= someShortIntervalInHours <= 6). I'm not sure the best way to choose that number but I will stick with 2-3 and change accordingly when we study traffic.

TheDanieLNK (changes requested):
See my comment on choosing lastReportDate. Otherwise, looks great and awesome job spotting the bug.

lib/fetching/content-services/crowd-tangle-content-service.js
mkyaw6: Thanks for the input Daniel, that makes sense. I have made the changes you have requested. Could you check the [change](https://github.com/TID-Lab/aggie/pull/336/commits/d7a33381dd95ff64e5701a56b00914532ff20c0d)?

TheDanieLNK (approved):
Looks good.

Issue #326, opened 2020-06-11
Slack & SES loggers should be disabled by default


karpawich:
Aggie uses [Winston ](https://github.com/winstonjs/winston)as a general-purpose logging framework. Logs at the `debug` level are written out to files, while `error` level logs & exceptions are written to a Slack channel and email address with AWS SES.

In development, the Slack and SES transports do not need to be in use unless they are being worked on directly. I propose that we allow them to be disabled via the `secrets.json` file, and that they are disabled by default.

Furthermore, I also propose that we introduce a simple `console` transport that developers can use in replacement of the Slack and SES transports for having immediate access to errors and unhandled exception stacktraces.

PR #337, opened 2020-07-14
Replace docker with more explicit setup scripts

cooperka:
For the most part, these scripts are verified working on a real production server! Since I wrote them as I went, we won't know whether I forgot anything small until after the first official setup using these steps ;)

Changes:
- Added `pm2` dependency for running in production
- Automatically disabled the Slack transport if it hasn't been configured (otherwise it fails), related to #326
- Replaced Docker scripts with shell scripts as discussed on Slack
    - These scripts are much more comprehensive than the Docker setup was, and result in a fully working production web server 🚀 
- Clarified and condensed other parts of the readme related to setup


sripushkar (changes requested):
The pm2 and the documentation looks good. I tested the added shell script to ensure it works as planned, but I am not on Ubuntu(I use Mint). I came across a couple errors:

In the 2nd step of installing aggie with the new shell script, when you use the command, $EDITOR config/secrets.json, it also puts the hashtagged comments at the top of the .json file, invalidating the syntax. 

I went in manually and fixed it, and then tried to run npm start, but I came across a error in one of the js files, which I attached below. I don't think this is a problem with the shell script, because when I tried to build the latest version of Aggie manually, I came to the same syntax error.

I don't think any of these issues were because of my Linux installation, but I think the main thing to fix is the part where it opens an editor for the secrets.json file, and added hashtagged lines to the file.

image

karpawich (changes requested):
When running the setup script on a fresh DigitalOcean Droplet, Ubuntu 18.04, the `nginx` service fails to restart because it uses the template config fetched from the Aggie repo. Upon running `systemctl status nginx.service` to investigate the failure, it shows the following error log:

```
invalid number of arguments in "ssl_certificate" directive in /etc/nginx/sites-enabled/aggie.conf:21
```

This makes perfect sense, considering that the value for the `ssl_certificate` directive is set to a template value:
```
<path to ssl certificate e.g. /etc/ssl/certs/cert.org.pem>
```

I tried reversing the order of the `nginx` service restart command and the editing of the `nginx` config file by the user, like so:
```
$EDITOR /etc/nginx/sites-available/aggie.conf
sudo systemctl restart nginx
```

such that `nginx` starts with a config file already filled in. But there does not appear to be any existing SSL certificate to point to - certainly not in the `/etc/ssl/certs` folder.

Perhaps additional steps should be introduced that facilitate the generation of an SSL certificate for `nginx` to start with?

cooperka:
Oh, dammit! That part will be hard to review because the file itself is modified in this PR. That config is no longer needed because of LetsEncrypt/certbot. For now you'll need to reference this branch:

```
sudo curl -o /etc/nginx/sites-available/aggie.conf https://raw.githubusercontent.com/TID-Lab/aggie/docker-removal/docs/content/aggie-nginx
```

README.md
```diff
@@ -40,84 +39,160 @@ Contact mikeb@cc.gatech.edu for more information on the Aggie project.
...
+
+sudo apt update
+sudo apt install -y ntp nginx software-properties-common
+sudo systemctl enable ntp
+sudo add-apt-repository universe
+sudo add-apt-repository ppa:certbot/certbot
+sudo apt update
+sudo apt install -y certbot python3-certbot-nginx
+
+sudo curl -o /etc/nginx/sites-available/aggie.conf https://raw.githubusercontent.com/TID-Lab/aggie/develop/docs/content/aggie-nginx
```
cooperka, line 114: Note: Until this PR is merged, use `docker-removal` instead of `develop` here.

karpawich: Looks like I got too excited :sweat_smile:, I'll swap in the right config file when I copy-paste the setup script over and run it :+1:

README.md
```diff
@@ -40,84 +39,160 @@ Contact mikeb@cc.gatech.edu for more information on the Aggie project.
...
+git clone https://github.com/TID-Lab/aggie.git
+cd aggie
+nvm install && npm install
+
+cp config/secrets.json.example config/secrets.json
+# User input: Customize Aggie settings per the README instructions.
+$EDITOR config/secrets.json
+
+# Ready! Test run:
+npm start
```
karpawich, line 161: The user should also start the `mongod` service to test Aggie, since it was only enabled & won't start until reboot.
```bash
sudo systemctl start mongod
npm start
sudo systemctl stop mongod
```

karpawich (commented):
**Update:** I reinstalled everything using the `docker_removal` branch, and it worked! All set :+1:

README.md
```diff
@@ -40,84 +39,160 @@ Contact mikeb@cc.gatech.edu for more information on the Aggie project.
...
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+```
+
+```shell script
+# Set up Aggie
+
+git clone https://github.com/TID-Lab/aggie.git
```
karpawich, line 152: Note: Until this PR is merged, use `docker-removal` instead of `develop` here, too!

cooperka:
Since this has now been verified by at least one installation, and since it only changes docs, I'm going to merge to develop to prevent the need for "check out this temporary branch" workarounds. Michael is using these steps to set up a new server as we speak.

I welcome more feedback, and will continue receiving notifications if you post messages in this PR. I'm sure there are ongoing improvements to be made to these instructions! Thanks all!

PR #339, opened 2020-07-14
Working substring for Author and Keyword

limmabean:
This does not work for Tags, but since tags are single phrases anyway, it shouldn't be hard to search the full tag.


karpawich:
[Search bar does not return substring matches, just word matches.](https://trello.com/c/Xe9EJet3/78-search-bar-does-not-return-substring-matches-just-word-matches)

models/query/report-query.js
```diff
@@ -68,15 +69,19 @@ ReportQuery.prototype.toMongooseFilter = function() {
   // Determine author filter
   if (this.author) {
     filter.author = {};
-    filter.author.$in = toRegexp.alli(this.author.trim().split(/\s*,\s*/).sort());
+    filter.author.$options = 'i';
+    filter.author.$regex = this.author;
```
cooperka, line 72: Would love a basic explanation of what this did before and what it does now. I'm curious why they were all turned into an array before if it's as simple as using $regex instead. Does this mean we can only search one author at a time now?

models/query/report-query.js
```diff
@@ -68,15 +69,19 @@ ReportQuery.prototype.toMongooseFilter = function() {
...
   }
   if (this.tags) {
-    filter.tags = { $all: toRegexp.allCaseInsensitive(this.tags) };
```
cooperka, line 74: Looks like the only remaining `toRegexp` usage after this is in `incident.js`, can we get rid of that too and delete the whole `toRegexp` file? Would be a nice cleanup, that file seems smelly to me.

cooperka (approved):

TheDanieLNK (approved):
Looks good to me. Thanks Matt.

models/query/report-query.js
limmabean: Previously we used a handwritten method of regexing the inputs. However, mongoDB now supports built in regex functions, so this is unecessary. Somewhere along the way, this previous method stopped working very well.

limmabean: Also yes, we can only search one author at a time.

limmabean:
Further Improvements include standardizing the search architecture between incidents and reports and supporting multiple searches in bars. This is currently limited because we don't use $or or $and functionality in our filters. This is not difficult to implement if we have a predetermined set of delimiters and indicators of $and and $or searches (eg. "Car + Tire" searches for posts with "Car" and "Tire")

PR #343, opened 2020-07-17
Fix shared posts

mkyaw6:
Frontend now displays text content from shared CrowdTangle posts. Shared content is displayed similarly to how Facebook displays shared text.


mkyaw6:
[Shared CrowdTangle Posts do not display any content](https://trello.com/c/973U22pP/74-shared-crowdtangle-posts-do-not-display-any-content)

limmabean (approved):

PR #346, opened 2020-07-21
fix: Make save() async to fix ParallelSaveError

cooperka:
Error when fetching well-formatted NEMO data: `Can't save() the same doc multiple times in parallel`

To verify:
- [ ] Does this break anything else?
- [ ] Does anything else also need to be made async? (due to mongoose upgrade perhaps?)


cooperka:
Verified again with latest code

PR #348, opened 2020-07-28
Aggie CT pagination fix

harshilshah4251:
- Crowdtangle posts are fetched using the Next Page URL in the paginated CT response. When the next page does not exist, Aggie makes a new call using the updated start date.

- Existing last report date functionality did not work with the paginated fetching. Updated it so that the last report date is always the max date (latest date) out of all the fetched posts.

Side note: I tested it myself multiple times but I would like any of you to test it out too because I think its a major change that could impact the number of posts we receive from CT.

### DIFF
--- a/lib/fetching/content-services/crowd-tangle-content-service.js
+++ b/lib/fetching/content-services/crowd-tangle-content-service.js
@@ -12,6 +12,7 @@ var _ = require('underscore');
 //options.lastReportDate is passed here through the content-service-factory and will be utilised in calling the api, but its actually only maintained by the parent content service, it is only utilised by the child
 var CrowdTangleContentService = function(options) {
 	this._baseUrl = config.get().crowdtangle.baseUrl;
+	this._nextPageUrl = undefined;
 	this._pathName = config.get().crowdtangle.pathName;
 	this._count = config.get().crowdtangle.count;
 	this._language = config.get().crowdtangle.language;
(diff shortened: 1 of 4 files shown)


lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -21,7 +22,7 @@ var CrowdTangleContentService = function(options) {
 	this._lastReportDate = options._lastReportDate;
 	this._listIds = parseInt(options.tags);
 	this.fetchType = 'pull';
-	this.interval = 30000;
+	this.interval = 1000;
```
mkyaw6, line 25: When testing the code out, aggie seems to be sending too many requests. Why was the interval decreased and could this be the reason it is erroring?
Screen Shot 2020-07-30 at 1 40 49 PM

mkyaw6: Resolved. A CT api key with increased rate limit needs to be used.

mkyaw6 (approved):
Looks good. Tested and works well.

TheDanieLNK (approved):
Looks good. Please confirm that 1000 is not a hard cutoff for the interval. We want to allow some random space for faster calls not to exceed the rate limit.

harshilshah4251:
> Looks good. Please confirm that 1000 is not a hard cutoff for the interval. We want to allow some random space for faster calls not to exceed the rate limit.

Yes. We get 100 calls/minute from CT. Interval of 1000 means 60 calls/minute. So we should be okay.

PR #350, opened 2020-07-28
Updates to Bootstrap 4.5 (with new date-time picker), Jquery 3.5.1, Moment 2.27, ui-bootstrap 0.12.0 and Media Display Features

limmabean:
Major overhauls in SCSS, templates, and angular controllers with a few changes in directives. This also includes media content additions (thumbnail and content). This also includes several updates that were made by Max to add API endpoints for SMTC tags.

This passes 219 test cases and fails 11. There are more failed tests in the develop branch (14).

This does not address a problem with will be fixed in a different branch: the show page of reports (when you click the content of a report) does not allow you to flag or add a report to an incident. This also does not include the tagging functionality. This will be accomplished in a different branch. 

I attempted to use BEM modifiers with all new CSS classes and remove as many uses of old classes as I could. The old CSS classes still largely remain, so some cleanup will be done soon after as well. If we could use BEM naming conventions from now on, it would be helpful.

[Added by Kevin:] BEM http://getbem.com/introduction/

The UI-bootstrap library has been modified and must be kept. This is because 0.12.0 does not support Bootstrap 4.5 but newer UI-bootstrap libraries require newer versions of angularjs. If we switch to angular 1.4 or higher, we can avoid using custom library files. 

There are still aria labels that need to be filled out, but there are no visual problems I can find.


limmabean:
I removed the SMTC commits for clairity.

cooperka (commented):
MASSIVE improvements 👏 

I tested for usability/bugs, and only glanced at the code changes which should have more thorough review by someone else. A couple thoughts (fixed!):
- ~~The statsbar floating on the right goes away entirely with smaller screen sizes; could it instead go at the top or bottom so it's not lost?~~
- ~~Incident view pages e.g. http://aggie.test/incidents/5f171dad1771033da8b7b174 seem to have wonky styles, they're the only page not consistent with everything else (the statsbar is rendered at the bottom and the spacing is weird)~~

public/angular/templates/reports/show.html
```diff
@@ -1,47 +1,242 @@
...
+                <td colspan="10" class="detail-value" ng-bind-html="report.content">{{ report.content }}</td>
+              </tr>
+              <tr ng-if="report.metadata.description">
+                <th scope="row" class="detail-key" translate>Shared Content</th>
+                <td colspan="10" class="detail-value" ng-bind-html="report.metadata.description">{{ report.metadata.description }}</td>
+              </tr>
+              <tr>
+                <th scope="row" class="detail-key" translate>Author</th>
+                <td colspan="10" class="detail-value">
+                  <a ng-href="{{report.metadata.accountUrl}}" target="_blank">{{ report.author }}</a>
```
cooperka: This link looks clickable but **does nothing** on click for me:

Screen Shot 2020-07-30 at 09 44 23

public/angular/templates/reports/show.html
```diff
@@ -1,47 +1,242 @@
...
+            <div class="right-side">
+              <div class="btn-group mr-4" role="group" aria-label="Flag/Unflag Buttons">
+                <button type="button" class="btn btn-info" ng-click="toggleFlagged(false)">
+                  <span class="mr-2" translate> Un-Flag </span>
+                  <i class="fa fa-flag-o" aria-hidden="true"></i>
+                </button>
+                <button type="button" class="btn btn-danger" ng-click="toggleFlagged(true)">
+                  <span class="mr-2" translate> Flag </span>
+                  <i class="fa fa-flag" aria-hidden="true"></i>
+                </button>
```
cooperka: Flag/unflag seems to have no effect here on the report show page

public/angular/templates/reports/show.html
limmabean: Fixed in https://github.com/TID-Lab/aggie/commit/899a45fee7d0c0358cd6288564c7fa60b37ecfe3. Also added views for twitter posts.

limmabean: > * Incident view pages e.g. http://aggie.test/incidents/5f171dad1771033da8b7b174 seem to have wonky styles, they're the only page not consistent with everything else (the statsbar is rendered at the bottom and the spacing is weird)

Fixed with https://github.com/TID-Lab/aggie/commit/40d320741d82f098f25e9864f9bea898f2ad7f13.

public/angular/templates/reports/show.html
limmabean: At the time of the PR they were placed there as placeholders because the functionality didn't actually exist on the original aggie. It is now done. See https://github.com/TID-Lab/aggie/commit/0b02c9e7fda490b0e673049feedd484364241a03 for the fix.

limmabean:
Toggling should be a single button like bold or italicize in word.

limmabean:
> * The statsbar floating on the right goes away entirely with smaller screen sizes; could it instead go at the top or bottom so it's not lost?

Addressed in https://github.com/TID-Lab/aggie/commit/ab6d7a404e0e91f175d817990b702d0fcdf3cf5c.

config/passphrase.txt
```diff
@@ -0,0 +1 @@
+32guns
```
mkyaw6, line 1: This seems to be the passphrase previously needed for PEM and should not be included in the PR

mkyaw6 (approved):

cooperka:
I pushed a tweak mainly updating docs; looks great!

PR #352, opened 2020-08-08
Refresh ct lists

mkyaw6:
Features added:

- Service to update crowdtangle_lists.json by calling the crowdtangle list api
- Button on frontend to call the service to update
- Cron job to call service at every UTC 00:00:00
- Backend test.


harshilshah4251 (commented):
Looks good to me overall. Just make sure whether or not Aggie needs to be restarted when the list is refreshed to account for new additions. If it needs to be restarted, then we can discuss how to achieve that by also loading the list in ct-content-service.js from memory rather than from the file.
Thanks.

TheDanieLNK (approved):
Looks great! Have you confirmed Harshil's question for both the button and cron refreshes?

mkyaw6:
> Looks good to me overall. Just make sure whether or not Aggie needs to be restarted when the list is refreshed to account for new additions. If it needs to be restarted, then we can discuss how to achieve that by also loading the list in ct-content-service.js from memory rather than from the file.
> Thanks.


Thanks for the catch. I've addressed this in the latest [commit](https://github.com/TID-Lab/aggie/pull/352/commits/b66c78aa1ba259c90d226b19466bb499bd571df6)

harshilshah4251 (approved):
LGTM

PR #353, opened 2020-08-11
Made sure to delete every listener that is created.

limmabean:
My hypothesis is that EventListeners are being created with socket.io and not being deleted. This obviously results in memory leaks on the part of the listener.


harshilshah4251 (approved):
LGTM

PR #355, opened 2020-08-13
Rename CT => FB in Aggie UI; remove old FB API

mkyaw6:


mkyaw6:
[Remove CrowdTangle from the UI (so trackers only see Facebook, Instagram, ...) and remove "old" Facebook stuff](https://trello.com/c/ei1CXcvU/100-remove-crowdtangle-from-the-ui-so-trackers-only-see-facebook-instagram-and-remove-old-facebook-stuff)

cooperka:
Nice! Can you also look under `docs/` and update the language there? That might be a good place to add a simple explanation that FB is accessed via CrowdTangle.

limmabean:
I think this needs to be more widespread. I think the removal of facebook from the entirety of the platform is more sensical. I'm working on doing that now.

cooperka:
The docs now use the word "CrowdTangle" and the UI uses the word "Facebook". I think this is fine for now. As noted a couple weeks ago, CT does more than just FB so eventually we should find a more generic way of presenting the CT option to users, but Aggie didn't support the other CT options before anyway so no worries.

PR #356, opened 2020-08-14
increased queue size

harshilshah4251:
Increasing the queue size. Max number of posts in CT's response is 100. 50 was too less to consume all 100 posts every second. 

So increased it to 200. Potentially, it will help avoid dropping posts when the queue is full.


limmabean (approved):
I could not find anything wrong with upping the default capacity.

PR #357, opened 2020-08-18
Table Fix

limmabean:
Added max-width to tags and location as these are potentially limitless.


cooperka:
Updated https://myanmaraggie.cartercenter.org/ and verified this fixes it ❤️

PR #359, opened 2020-08-26
Fix Source Creation Modal for Crowdtangle

limmabean:
This simply hides the keywords field when CT is the source type.


limmabean:
Also changed the wording of the source table because it's inconsistent with the fields in the modal.

harshilshah4251 (approved):
lgtm

PR #360, opened 2020-08-26
Saved search fix

harshilshah4251:
- Adding support for getting posts from Saved Searches on CT

- Aggie alternates between requesting posts from lists and posts from saved searches. This had to be done because CT won't include posts from saved searches in the normal /posts endpoint without list ids. 

- Changed the script that makes CT list <->acct pairs to also include information on saved searches.

### DIFF
--- a/config/crowdtangle_list.json
+++ b/config/crowdtangle_list.json
@@ -1,4 +1,0 @@
-{
-	"crowdtangle_list_account_pairs": {
-	}
-}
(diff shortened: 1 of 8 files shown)


limmabean:
Remove CTLists Json files. They are not for public view.

limmabean (approved):
I can only confirm this works on my device. I have no good way to test this. However, the added code looks correct and well written.

PR #362, opened 2020-09-01
Resolve security issue with refreshing CTLists by hardcoding output directory

RoboticWater:
https://trello.com/c/nmnr8Mg0/112-improve-security-for-user-input-directory-when-refreshing-ct-lists

Currently the `/api/v1/settings/updateCTList/:directory` endpoint allows requesters to specify the output directory of the CTList output, but not only is this unsafe for reasons described by the trello issue, it's not really necessary. The only reason we can't hardcode the directory is that the tests need to specify a custom output directory. 

I've solved this by splitting the endpoint into two: one exposed during production and development that relies on a hardcoded output directory, and one for testing that has the output directory as a parameter.

I'm not sure my solution is ideal though. Maybe there's a way to run the tests without specifying a custom output directory.


lib/api/CT-list-update-service.js
```diff
@@ -9,7 +9,7 @@ var CTListUpdateService = function(options) {
         'Cache-Control': 'no-cache'
     }
     this._apiToken = config.get().crowdtangle.apiToken;
-    this._outputDir = options.directory;
+    this._outputDir = options ? options.directory : '../../config/crowdtangle_list.json';
```
cooperka: Simplification: Instead of dynamically assigning this hardcoded string here, you could use below `CTListUpdateService.prototype.OUTPUT_DIR = '../../config/crowdtangle_list.json'` to make it static.

Then in the test file, you could import this class and change the static variable in the `stubWithFixture` method that already exists: something like `CTListUpdateService.prototype.OUTPUT_DIR = outputFixtureFile`.

This will allow us to use nearly the same logic in test and production, just with a different variable name (instead of using an entirely different test route which could lead to unintended bugs later on if we forget to keep them in sync).

lib/api/v1/settings-controller.js
```diff
@@ -45,8 +45,17 @@ app.put('/api/v1/settings/updateCTList/:directory', async function(req, res) {
...
+} 
+app.put('/api/v1/settings/updateCTList', async function(req, res) {
+  updateCTList(req, res, new CTListUpdateService());
 });
 
+if (process.env.NODE_ENV === 'test') {
+  app.put('/api/v1/settings/updateCTList/:directory', async function(req, res) {
+    updateCTList(req, res, new CTListUpdateService({directory: req.params.directory}));
+  });
+}
```
cooperka: This will be no longer needed

cooperka (commented):
Lovely work! I gave some feedback about improving it further, let me know if it makes sense.

cooperka (approved):
👍 💯 ❤️

PR #365, opened 2020-09-08
Fixed error in security update to CT List output directory

RoboticWater:
When I merged the previous fix #362, it had overwritten a line and forgot to update a variable name. I have resolved that here.

This version still fails the last two tests in `test/backend/lib.api.CT-list-update-service.test.js`, but that appears to be a problem with something else, not my updates. The last test specifically seems to fail because the custom `_parseList()` function that is used for the test structures its return value in an unexpected way.


cooperka (approved):
I think GitHub wanted to re-merge **everything** in your branch (not just the fix) so I rebased and only kept the very last commit. All it does now is change that one variable. If that looks right to you, please go ahead and merge! 👍

PR #366, opened 2020-09-15
Fix GO button, change CT-> FB, page/groups-lists mapping change

harshilshah4251:
1.  Change 1-1 mapping (pages/groups -> list) to 1-many ( pages/groups -> [list1, list2, ...] )
-  This change was made because certain pages/groups exist in multiple lists and it's important to show all lists on Aggie Frontend. 

2. Change CT -> Facebook everywhere

3. Fix the GO button.


public/angular/templates/reports/index.html
```diff
@@ -25,7 +25,7 @@
               <div class="form-group col mb-0">
                 <div class="input-group input-group-sm">
                   <input type="text" ng-model="searchParams.tags" aria-label="tagSearch" class="form-control filters__input" placeholder="{{'Enter Tags' | translate}}">
-                  <div class="input-group-append" ng-click="clearTags()">
+                  <div class="input-group-append">
```
cooperka, line 28: Is this the actual change that fixed the Go button? Simple!

lib/fetching/bot.js
```diff
@@ -102,7 +102,7 @@ Bot.prototype._reportListener = function(reportData) {
       if (reportData._media === 'twitter') {
         reportData.metadata.retweet ? reportData.tags.push('RT') : reportData.tags.push('NO_RT');
       }
-      if (reportData._media === 'crowdtangle') {
+      if (reportData._media === 'facebook') {
```
cooperka, line 105: Is this a user-facing string or just internal? I personally think it's more accurate to use "crowdtangle" for internal things because that's what it truly is. We just want to use "facebook" when showing a string to the trackers, because they don't actually know or care what crowdtangle is.

In any case, this doesn't seem to fix the particular string here; I still see "__CROWDTANGLE" unless maybe I need to reconfigure something in settings:

Screen Shot 2020-09-08 at 11 12 21

cooperka (approved):
It works for me, and looks good aside from a preference of not changing "crowdtangle" internally. Feel free to discuss. I'm happy merging this either way.

lib/fetching/bot.js
harshilshah4251: Did you try doing `node install.js && gulp angular`? I think thats needed when front end changes are made.

public/angular/templates/reports/index.html
harshilshah4251: Yea. Loll

lib/fetching/bot.js
cooperka: Ah, I did something similar `npm install` (which triggers `gulp build`) and that resolved it. Seems like `npx gulp angular` is faster, I didn't know about that one. None of this should be necessary in development but unfortunately it is for now 🙃

PR #368, opened 2020-09-22
Fix RSS

cooperka:
Fixes non-CT sources by allowing both methods of returning data (see code comment for more info).

Example RSS feed to test: https://www.nasa.gov/rss/dyn/breaking_news.rss


harshilshah4251 (approved):
LGTM

PR #369, opened 2020-09-24
fixed ct-update button, CT-> FB (only frontend).

harshilshah4251:
CT-update list button:

ReloadCTList threw an error because it was being invoked by bot.contentservice.reloadCTList. When the contentservice was anything other than Crowdtangle, using this functionality would throw an error because other contentservices do not have reloadCTList function.

Hence, added a conditional to ensure that the contentservice is CT before invoking reloadCTList. 

Reverted backend changes from CT-> FB to FB-> CT (as it was earlier). Only changed the frontend to show FB instead of CT.

### DIFF
--- a/public/angular/js/filters/replaceCT.js
+++ b/public/angular/js/filters/replaceCT.js
@@ -1,0 +1,14 @@
+angular.module('Aggie')
+// Facebook is the source. Crowdtangle is the API fetching content from Facebook. 
+// Frontend users only concerned with the source(Facebook). 
+// replaceCT replaces frontend references to Crowdtangle with Facebook.
+.filter('replaceCT', function() {
+  return function(string) {
+    if (!string) return;
+    if (string === 'crowdtangle') {
+      return 'facebook';
+    } else {
+      return string;
+    }
+  };
+});
--- a/public/angular/templates/api-settings-modal.html
+++ b/public/angular/templates/api-settings-modal.html
@@ -14,12 +14,6 @@
         <li translate>api_settings_modal.twitter.p1</li>
         <li translate>api_settings_modal.twitter.p2</li>
       </ol>
-      <ol ng-if="item === 'facebook'">
-        <li translate>api_settings_modal.facebook.p1</li>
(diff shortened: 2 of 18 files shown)


public/angular/js/filters/replaceCT.js
```diff
@@ -0,0 +1,12 @@
+angular.module('Aggie')
+
+.filter('replaceCT', function() {
```
cooperka, line 5: Neat! Could you please just add a code comment here explaining WHY this is being done? Future devs will probably scratch their heads at this 🙃

public/angular/templates/api-settings-modal.html
```diff
@@ -14,12 +14,6 @@ <h5 class="modal-title">
         <li translate>api_settings_modal.twitter.p1</li>
         <li translate>api_settings_modal.twitter.p2</li>
       </ol>
-      <ol ng-if="item === 'facebook'">
-        <li translate>api_settings_modal.facebook.p1</li>
-        <li translate>api_settings_modal.facebook.p2</li>
-        <li translate>api_settings_modal.facebook.p3</li>
-        <li translate>api_settings_modal.facebook.p4</li>
-      </ol>
```
cooperka, line 22: Clearly the old Facebook help was obsolete here, but might we replace it with CT help instead of removing it entirely?

cooperka (commented):
Looks good, just a question and a tweak. I have not tested this in the actual app, just looked at the code.

public/angular/js/filters/replaceCT.js
harshilshah4251: sure

public/angular/templates/api-settings-modal.html
harshilshah4251: done!

limmabean:
LGTM

limmabean (approved):
Looks good to me

cooperka (approved):
While testing I fixed a couple regressions I found that were unrelated to your changes here. This looks good and seems to work. Merging!

PR #371, opened 2020-09-24
CT List dropdown

RoboticWater:
Adds a dropdown to filter by Crowd Tangle list.
- Added another route so the client can query the `crowdtangle_list.json`. I'm not sure if this is overkill. Given that the list can get updated, I assumed clients would need a way to query it on the fly. 
- Changed the query for reports that assumes `report.tags` is an array. This was necessary to get the filter to work. 
- Changed the way that the `ct_tag` is stored in the server to keep it as an array. I updated the front end code to accommodate both the new format and the old.

We might want to make the dropdown only visible if media type or source is set to facebook, but we might not have to.


config/crowdtangle_list.json
```diff
@@ -1,4 +1,924 @@
 {
-	"crowdtangle_list_account_pairs": {},
-	"crowdtangle_saved_searches": {}
+	"crowdtangle_list_account_pairs": {
```
harshilshah4251: Edit the content of this file to only show the following template. Not useful to show list names from the dashboard we are using.
{
	"crowdtangle_list_account_pairs": {},
	"crowdtangle_saved_searches": {}
}

lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -182,7 +182,7 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
 	this.crowdtangle_saved_searches = crowdtangle_lists.crowdtangle_saved_searches;
 	// If the list is found and matched, then the ct_tag is the list name
 	if (this.crowdtangle_lists[data.account.id]) {
-		metadata.ct_tag = this.crowdtangle_lists[data.account.id].join(", ");
+		metadata.ct_tag = this.crowdtangle_lists[data.account.id];
```
harshilshah4251, line 185: Does this mean that a post is only associated to 1 list? I think we want a post to have multiple lists. Post x can be in page Y that exists in two different lists. Hence, post x would also be in two different lists.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"editor.formatOnSave": false
```
harshilshah4251: add this to .gitignore

lib/fetching/content-services/crowd-tangle-content-service.js
RoboticWater: No, this just means that the datastructure will be an array not a string, so if a report has lists LIST_A and LIST_B, it'll look like `["LIST_A", "LIST_B"]` not `"LIST_A, LIST_B"`. I had to change the front end to accommodate this, but it lets us search by each list more naturally with the `$in` query rather than a regex search.

config/crowdtangle_list.json
RoboticWater: Should we add this to the .gitignore? It changes whenever you update the CTLists

lib/fetching/content-services/crowd-tangle-content-service.js
harshilshah4251: cool

harshilshah4251 (approved):
LGTM

limmabean (approved):
LGTM

PR #372, opened 2020-09-29
Fixed CTLists not working in batch mode

RoboticWater:
I forgot to add the CTList provider to the batch route. This has been rectified


limmabean (approved):
It works now. Good to go.

PR #374, opened 2020-09-30
fixed pagination incident search

harshilshah4251:
I made the pagination function in incidents/index.js the same as the pagination function in reports/incident.js.

I think items.slice was causing the problem and returning only 25 items that match the search. Removing it fixed the issue. But I am not sure why it was added in the first place.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -376,7 +376,7 @@
     "JSONStream": {
       "version": "1.3.5",
       "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
-      "integrity": "sha512-E+iruNOY8VV9s4JEbe1aNEm6MiszPRr/UfcHMz0TQh1BXSxHK+ASV1R6W4HpjBhSeS+54PIsAMCBmwD06LLsqQ==",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -376,7 +376,7 @@
     "JSONStream": {
       "version": "1.3.5",
       "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
-      "integrity": "sha512-E+iruNOY8VV9s4JEbe1aNEm6MiszPRr/UfcHMz0TQh1BXSxHK+ASV1R6W4HpjBhSeS+54PIsAMCBmwD06LLsqQ==",
+      "integrity": "sha1-MgjB8I06TZkmGrZPkjArwV4RHKA=",
```
cooperka: @harshilshah4251 could you please run `npm install -g npm` to get the latest version on your machine? Then these hashes won't keep flipping.

cooperka (approved):
Tested and works great. I also have no idea what led to the original change (commit history is extremely vague) but it's now consistent with the reports page so I see no problem with it. Thanks!

PR #375, opened 2020-10-06
SMTC Tags Auto Complete Search

limmabean:
Currently only $all search is supported. This mean if I search two tags together, I will get all the posts that contain those two tags. Those posts can also contain other tags, but as long as they have the searched tags, they will appear. If you search something that isn't a tag, it will be removed before searching.


harshilshah4251:
Tested it and it works as expected! Approved

harshilshah4251 (approved):

RoboticWater:
UX Concerns
- When a user completes a tag search, the tags momentarily convert to the tag ids before the next page loads; might be confusing 
- URL querystring currently uses tag ids rather than names, which might be confusing for users; may also be inconvenient if  users like to manipulate the url directly or look at old searches in history
- I'm used to hitting _either_ tab _or_ enter to complete an autocomplete, but here tab just moves to the next input like default; not sure if the average user will care though

Functionality seems to be there though.

RoboticWater (approved):

PR #376, opened 2020-10-07
Determining saved search names using substring search

harshilshah4251:
image
Adding saved search names to each post using substring search. 

Also, I accidently pushed Twitter Hate Speech Api commit here. Is it okay if you can review that as well here? I will delete the other PR.
Adding Daniel for Twitter HS API.
Thanks.

### DIFF
--- a/lib/api/CT-list-update-service.js
+++ b/lib/api/CT-list-update-service.js
@@ -13,7 +13,8 @@ var CTListUpdateService = function() {
 }
 CTListUpdateService.prototype.OUTPUT_DIR = '../../config/crowdtangle_list.json';
 CTListUpdateService.prototype._updateCTList = async function() {
-    var content = { "crowdtangle_list_account_pairs": {}, "crowdtangle_saved_searches" : {}};
+    var content_raw = fs.readFileSync(path.resolve(__dirname, this.OUTPUT_DIR));
+    var content = JSON.parse(content_raw);
     await this._fetchLists();
     if (this._listResponse) {
         // build array of objects with list information
@@ -49,7 +50,7 @@ CTListUpdateService.prototype._updateCTList = async function() {
                     content.crowdtangle_list_account_pairs[account.id] = [list.metadata.title];
                 } 
                 else {
-                    content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title);
+                    content.crowdtangle_list_account_pairs[account.id].indexOf(list.metadata.title) === -1 && content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title); //if listname is not in the array, then it will be added. 
                 }
                 
             });
@@ -57,7 +58,10 @@ CTListUpdateService.prototype._updateCTList = async function() {
         
         while(savedSearches.length !== 0) {
             var savedSearch = savedSearches.shift();
-            content.crowdtangle_saved_searches[savedSearch.list_id] = savedSearch.title;
+            if(!(savedSearch.list_id in content.crowdtangle_saved_searches)){
+                content.crowdtangle_saved_searches[savedSearch.list_id] = {"name" : savedSearch.title, "words": []};
+            }
+            
         }
     }
     // writes to JSON
--- a/lib/api/CT-list-update-service.js
+++ b/lib/api/CT-list-update-service.js
@@ -50,9 +50,10 @@ CTListUpdateService.prototype._updateCTList = async function() {
                     content.crowdtangle_list_account_pairs[account.id] = [list.metadata.title];
                 } 
                 else {
-                    content.crowdtangle_list_account_pairs[account.id].indexOf(list.metadata.title) === -1 && content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title); //if listname is not in the array, then it will be added. 
+                    if (content.crowdtangle_list_account_pairs[account.id].indexOf(list.metadata.title) === -1){
+                        content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title);
+                    }
                 }
-                
             });
         };
         
--- a/lib/fetching/content-services/crowd-tangle-content-service.js
+++ b/lib/fetching/content-services/crowd-tangle-content-service.js
@@ -156,6 +156,7 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
 		sponsor: data.brandedContentSponsor	|| null,
 		caption: data.caption || null,
 		description: data.description || null, // this stores the content of sharedposts
+		imageText: data.imageText || null,
 		title: data.title || null,
 		crowdtangleId: data.id || null,
 		externalUrl: data.link || null,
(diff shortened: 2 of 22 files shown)


lib/api/CT-list-update-service.js
```diff
@@ -49,15 +50,18 @@ CTListUpdateService.prototype._updateCTList = async function() {
                     content.crowdtangle_list_account_pairs[account.id] = [list.metadata.title];
                 } 
                 else {
-                    content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title);
+                    content.crowdtangle_list_account_pairs[account.id].indexOf(list.metadata.title) === -1 && content.crowdtangle_list_account_pairs[account.id].push(list.metadata.title); //if listname is not in the array, then it will be added.
```
cooperka: This line got complex! Can we please use `else if (condition) { stuff }` instead of `else { condition && stuff }`

lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -184,7 +185,20 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
 	if (this.crowdtangle_lists[data.account.id]) {
 		metadata.ct_tag = this.crowdtangle_lists[data.account.id];
 	} else if (options.requestType == "SAVED_SEARCH") {
-		metadata.ct_tag = "Saved Search";
+		metadata.ct_tag = []
+		for(var saved_search_id in this.crowdtangle_saved_searches){
```
cooperka, line 190: Clever workaround, I'm glad this works! For the sake of others reading this later, could you please add a comment here explaining that this logic searches for a match to determine if it's likely related to a saved search, and also note that it's a workaround that we hope will go away once CT supports it?

lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -184,7 +185,20 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
...
+			var saved_search = this.crowdtangle_saved_searches[saved_search_id];
+			var regex = new RegExp(saved_search.words.join("|"), 'gi');
+			if(content.search(regex) >= 0 || 
+				(metadata.caption !== null && metadata.caption.search(regex) >=0) || 
+				(metadata.description !== null && metadata.description.search(regex) >= 0) ||
+				(metadata.imageText !== null && metadata.imageText.search(regex) >= 0)) {
```
cooperka: Tip: Instead of `if not null && string.search(regex)` you could simplify to `regex.test(string)` if you want to. No null check needed that way.

lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -184,7 +185,20 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
 	if (this.crowdtangle_lists[data.account.id]) {
 		metadata.ct_tag = this.crowdtangle_lists[data.account.id];
 	} else if (options.requestType == "SAVED_SEARCH") {
-		metadata.ct_tag = "Saved Search";
+		metadata.ct_tag = []
+		for(var saved_search_id in this.crowdtangle_saved_searches){
+			var saved_search = this.crowdtangle_saved_searches[saved_search_id];
+			var regex = new RegExp(saved_search.words.join("|"), 'gi');
```
cooperka: This will throw an error if saved_search.words contains characters like `(` that make the regex invalid. It's rare but we should at least catch the error here.

public/angular/js/controllers/reports/index.js
```diff
@@ -39,7 +39,9 @@ angular.module('Aggie')
     $scope.statusOptions = statusOptions;
     $scope.currentPath = $rootScope.$state.current.name;
     $scope.smtcTags = smtcTags;
-    $scope.listOptions = Array.from(new Set(Object.values(ctLists.crowdtangle_list_account_pairs).flat()));
+    $scope.listOptions = Array.from(new Set(Object.values(ctLists.crowdtangle_list_account_pairs).flat())).concat(Array.from(new Set(Object.values(ctLists.crowdtangle_saved_searches).map(function(obj) {
+      return obj.name;
+    }))), ["Saved Search"]);
```
cooperka, line 48: I know you're just copying logic that was already here, worth simplifying though -- I think we can just do `Object.values({})` in both cases instead of `Array.from(new Set(Object.values({})))` since it returns the same result (or maybe I'm misunderstanding something).

cooperka (commented):
Thanks for this important feature! I have mostly cleanup suggestions and a bug. I know it's an old codebase but worth keeping things tidy so it doesn't get worse.

There's also a merge conflict fyi (it looks like there are some changes that got pulled in here that are already on develop, `git rebase develop` may help).

public/angular/js/controllers/reports/index.js
harshilshah4251: There might be duplicates in the returned array after `flat` operation. So using `new Set` would only keep the unique values which is what we want. I think this is fine unless theres a better way to find unique values and make an array.

cooperka: Ah okay, yes there are easier ways with Lodash for example but not worth blocking this. Thanks for clarifying what's going on.

lib/fetching/content-services/crowd-tangle-content-service.js
```diff
@@ -184,7 +185,28 @@ CrowdTangleContentService.prototype._parse = function(data, options) {
...
+		// We plan to remove this workaround when CT adds list IDs/account IDs to saved search API responses.  
+		for(var saved_search_id in this.crowdtangle_saved_searches){
+			var saved_search = this.crowdtangle_saved_searches[saved_search_id];
+			try{
+				var regex = new RegExp(saved_search.words.join("|"), 'gi');
+			} catch (err) {
+				console.log("Error creating regex for saved searches");
+			}
+			if (saved_search.words.length && 
+				(regex.test(content)||
```
cooperka: Can you please test this with invalid content e.g. "foo(" and make sure it works as expected? (in that case it should fall back to generic "Saved search")

I think if we catch the error above, now `regex` will still be null and thus `regex.test` will throw.

cooperka: Fwiw, logic like this is often easier to deal with if you modularize it into a generic function. For example something like:

```js
const doesMatch = doesReportMatchWords(saved_search.words, content, metadata)
metadata.ct_tag.push(doesMatch ? saved_search.name : "Saved Search")
```

public/angular/js/controllers/reports/index.js
RoboticWater: Yeah, I wrote the original line there. I think I just didn't know we had lodash at the time.

We might want to add a note to rewrite this, because it's definitely verbose, if functional.

lib/fetching/content-services/crowd-tangle-content-service.js
harshilshah4251: Good catch. I moved the test statement inside the try block.

harshilshah4251: And tested it with that example. It produces an error and then it will fall back to Saved Search.

harshilshah4251: Yes. Modularized it. Moved the regex.test part inside the try block. "foo(" throws an error and we would catch it. If no saved search matches are found, then we would add Saved Search default tag.

cooperka (approved):
Fantastic! Thanks for the important cleanup. Code looks mergeable to me, but I have not tested the app itself. Feel free to merge if you've tested the full integration locally.

PR #378, opened 2020-10-09
Saved search Edge Case Fix

harshilshah4251:
When there are no saved searches, aggie threw some weird error. Handling that edge case in this PR.


limmabean (approved):
Can't confirm this fixes it because I'm struggling to recreate it, but this definetly does not break the platform.

RoboticWater (approved):

PR #379, opened 2020-10-10
Sorting By HateScoreSpeech

MrMathur:
Sorts the batch view mode in descending order of hateSpeechScore of every report.

Added a file './public/templates/reports/tableBatch.html' that contains the logic for sorting by the metadata.hateSpeechScore

Made changes to the file './public/templates/reports/batch.html' to call the new table. Also added help text to convey that the reports are sorted by HateSpeechScore.


harshilshah4251:
image


I tested it and I can see that it works. But I dont see any help text on the batch mode anywhere. Can you please upload a screenshot with that help text?

harshilshah4251 (approved):
Please address the comment about help text. Everthing else looks good to me!

harshilshah4251 (approved):

limmabean (approved):
Works well! Nothing about it could break tbh.

PR #380, opened 2020-10-13
Incident Modal Text Wrap Fix 

limmabean:
Fixed text exiting columns in the add report to incident modal. Also added an upper limit to the width the incident column can take given a large incident title.

### DIFF
--- a/public/angular/templates/reports/table.html
+++ b/public/angular/templates/reports/table.html
@@ -19,7 +19,7 @@
           </div>
         </td>
         <td class="td__sourceInfo">
-          <span class="badge badge-pill badge-secondary mb-2">{{ r.storedAt | interval }} ago</span>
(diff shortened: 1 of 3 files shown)


public/angular/templates/reports/table.html
```diff
@@ -19,7 +19,7 @@
           </div>
         </td>
         <td class="td__sourceInfo">
-          <span class="badge badge-pill badge-secondary mb-2">{{ r.storedAt | interval }} ago</span>
+          <span class="badge badge-pill badge-secondary mb-2" title="{{ r.storedAt | aggieDate:'date' }} at {{ r.storedAt | aggieDate:'short_time' }}">{{ r.storedAt | interval }} ago</span>
```
cooperka, line 22: Mouseover title is also a big improvement!

cooperka (approved):
Verified bonafide

PR #382, opened 2020-10-15
Ml flag and read toggle button and Other Visual Updates

limmabean:
Just wanted to do some extra passes to make small changes for consistency along the entire platform.

### DIFF
--- a/public/angular/js/controllers/reports/index.js
+++ b/public/angular/js/controllers/reports/index.js
@@ -408,19 +408,26 @@ angular.module('Aggie')
      * Sets selected reports' read property to read value
      * @param {boolean} read
      */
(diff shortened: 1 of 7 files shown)


RoboticWater:
Not sure how extensive these visual tweaks are supposed to be, but for the sake of consistency, you might want to make incident checkboxes bigger like the reports page: 
image

RoboticWater:
Might also want to do the same with these edit/delete icons:
image

Actually, I would add as much space as reasonably possible between these; it's easy to accidentally click delete when you mean edit (thankfully, it asks before you proceed, but it's still annoying).

limmabean:
Fixed John's Comments

RoboticWater (approved):

PR #383, opened 2020-10-15
SMTCTags contain Boolean field for enabling Comments

limmabean:
This is both the front-end and the back-end changes necessary. The field is called isCommentTag and defaults to false.


public/angular/templates/tags/modal.html
```diff
@@ -27,6 +27,10 @@ <h5 class="modal-title">
             A color must be specified.
           </small>
         </div>
+        <div class="form-group">
+          <label for="tagIsCommentTag" translate>Does tag enable comments?</label>
```
RoboticWater: Might want to rephrase this to better explain what it does. Like "Does this tag enable comment acquisition on posts?" or something. Might be fine though.

public/angular/templates/tags/table.html
```diff
@@ -5,6 +5,7 @@
     <th scope="col" translate>Color</th>
     <th scope="col" translate ng-if="currentUser.can('view other users')">Created By</th>
     <th scope="col" translate>Description</th>
+    <th scope="col" translate>Tag Enables Comments?</th>
```
RoboticWater: Same as above. Something like "tag enables comment acquisition on posts"

karpawich (approved):

PR #386, opened 2020-10-16
Data Visualization

RoboticWater:
Data visualization MVP. It requires that you have gotten data within the past 2 days and have tags (ideally pulled from over several hours), though you should make sure that I've captured those edge cases. It pulls data directly from the database, so we should make sure the endpoint is secure.

The code is pretty ugly, so I'm sorry.


config/crowdtangle_list.json
```diff
@@ -1,4 +1,921 @@
 {
-	"crowdtangle_list_account_pairs": {},
-	"crowdtangle_saved_searches": {}
+	"crowdtangle_list_account_pairs": {
+		"296278": [
```
cooperka: Please revert these changes (you can `git checkout develop config/crowdtangle_list.json` to accomplish this)

public/angular/js/services/visualization.js
```diff
@@ -0,0 +1,15 @@
+angular.module('Aggie')
+
+.factory('Visualization', function($resource) {
+	var endDate = new Date();
+	var startDate = new Date();
+	startDate.setTime(startDate.getTime() - 2 * (24 * 60 * 60 * 1000));
+	console.log("/api/v1/viz/?before=" + endDate.toISOString() + "&after=" + startDate.toISOString());
+  return $resource(
```
cooperka: There are some big spacing inconsistencies here, please convert all these tabs to spaces for consistency with the rest of the codebase (most editors allow you to set a default and easily change between them; you could also find/replace and do it all in one go).

Tabs tend to be problematic when working in teams because they render differently for everyone. Thank you!

public/angular/index.html
```diff
@@ -43,6 +43,7 @@
         <script src="/js/Autolinker.min.js"></script>
         <script src="/js/canvasjs.min.js"></script>
         <script src="/js/app.min.js"></script>
+        <script src="https://d3js.org/d3.v6.min.js"></script>
```
cooperka, line 46: A more conventional place to put this would be in package.json via `npm install --save d3` then import `d3` in files explicitly rather than using a global constant.

models/query/report-query.js
```diff
@@ -55,6 +55,7 @@ ReportQuery.prototype.toMongooseFilter = function() {
   if (this.keywords)  filter.content = { $regex: this.keywords,  $options: 'i' }
   if (this.tags)      filter.smtcTags = { $all: this.tags }
   if (this.list)      filter["metadata.ct_tag"] = {$in: [this.list] }
+  console.log(filter);
```
cooperka: Please remove, it logs quite frequently 😀

cooperka (commented):
I only had time for a quick glance at general style, so I didn't look at code too closely. This is awesome work! I'm really excited to have a working Analysis page again.

Have you tested it with RSS sources? For example https://www.nasa.gov/rss/dyn/breaking_news.rss => I am seeing blank graphs after adding several tags to this data, and I wonder if RSS data is somehow breaking it.

Screen Shot 2020-10-15 at 22 09 23

public/angular/index.html
RoboticWater: Yeah, I think I tried that initially, and it didn't work for whatever reason, so I just said screw it, we'll go with the CDN for the sake of getting something that works. But I'll try it again.

models/query/report-query.js
RoboticWater: Whoops; thought I got all of these.

config/crowdtangle_list.json
limmabean: Seconded please remove this file.

public/angular/index.html
RoboticWater: OK, yeah, the reason was that the d3 library doesn't play well with our build system. I guess it uses newer syntax or something, but you'll get this error if you don't use the CDN:
```
[11:44:04] Using gulpfile ~/Documents/web/aggie/gulpfile.js
[11:44:04] Starting 'build'...
[11:44:04] Starting 'debugTranslations'...
[11:44:04] Starting 'angular'...
[11:44:04] Finished 'debugTranslations' after 65 ms
[11:44:07] Plumber found unhandled error:
 Error: StringMap expected string key
[11:44:07] Finished 'angular' after 2.98 s
[11:44:07] Finished 'build' after 2.98 s```

RoboticWater:
> I only had time for a quick glance at general style, so I didn't look at code too closely. This is awesome work! I'm really excited to have a working Analysis page again.
> 
> Have you tested it with RSS sources? For example https://www.nasa.gov/rss/dyn/breaking_news.rss => I am seeing blank graphs after adding several tags to this data, and I wonder if RSS data is somehow breaking it.
> 
> Screen Shot 2020-10-15 at 22 09 23

Strange. It should work with any kind of report because it just uses the default report structure. Are RSS sources so radically different?

The only thing I think could be wrong here is that RSS sources don't have an `authoredAt` field.

public/angular/js/services/visualization.js
RoboticWater: This should be fixed now

harshilshah4251 (approved):
Good work! If you have some time, I would love to see at least the number of Hate Speech Posts somewhere on the dashboard. You will just need to count the number of reports with metadata.hateSpeechScore > 0.5 and display that number somewhere. Thanks!

And since there are so many commits, it might be better to squash and merge on develop.

public/angular/index.html
cooperka: I think you're right, seems like modern versions of D3 don't play well with the angular transpiler we're using (https://stackoverflow.com/a/45980599/763231 for example). This is fine for now then, thanks for the context.

limmabean (approved):
Went through the files. LGTM

RoboticWater:
@limmabean if you have time it'd be better if you could run it locally and see if it correctly renders the data. Again, only if you have time though.

PR #387, opened 2020-10-16
Consistency and Accessibility Updates and Removal of Individual Tag Edit

limmabean:
Did several expert reviews to improve consistency and accessibility

Removed Individual Tag Edit because comments acquisition breaks when having individual edits. This is a TODO item and has been clearly marked in the code as a TODO item.

### DIFF
--- a/public/angular/index.html
+++ b/public/angular/index.html
@@ -43,6 +43,12 @@
         <script src="/js/Autolinker.min.js"></script>
         <script src="/js/canvasjs.min.js"></script>
         <script src="/js/app.min.js"></script>
+        <script>
+          //Enable Tooltips Everywhere
+          $(function () {
+            $('[data-toggle="tooltip"]').tooltip()
+          })
+        </script>
       </div>
       <footer>
         <div class="version">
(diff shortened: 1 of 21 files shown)


RoboticWater:
I'm not sure its worth losing the current tag editing flow for this functionality. If trackers have more than around 7 tags (the location tags alone are more than that), then trackers will likely have difficulty remembering which tags are where. And they specifically requested autocomplete for the searching, so to lose it on the adding phase will not only be confusing, but definitely not what they want.

Really, it comes down to how much we need this functionality, but I feel like we might want to do it some other way. Could we queue comments via flagging instead?

cooperka:
@RoboticWater I will ask about this and make sure whatever we ship meets their needs. The tag modal feature is a bit complex to maintain (not to mention slow in terms of human interaction) so I see removing it as a win overall. Let's see where we get with this and we can always come back to it if needed.

I think a "search filter" on the tags list may be a better direction if we need this functionality (we can consider this soon).

RoboticWater:
@cooperka I agree, a filter on the tag sidebar would probably alleviate my concerns. The current flow still seems unintuitive, in that I don't think any other app does tagging like this, but since trackers are employees, we can just explain to them the process. The biggest issue is just being able to find the right tag.

cooperka:
@limmabean does this supersede https://github.com/TID-Lab/aggie/pull/335 which we can now close?

cooperka (approved):
So much consistency! I tested locally pretty exhaustively, seems to work great.

PR #389, opened 2020-10-16
adding list and tags because they are needed by socket-handler

harshilshah4251:
New reports showing in the banner even when they are not there. This PR fixes it.


limmabean:
Tested, Works Great.

limmabean (approved):
WORKS GREAT

PR #391, opened 2020-10-16
Default CT useLanguage to false

cooperka:
The idea is, instead of manually commenting out code on each machine, we can just default `useLanguage` to `false`. Then we don't need to maintain a git diff and stash/pop every time we deploy. This PR won't even require a config change on any machine since it's off by default.

I've tested this locally, just want to have @harshilshah4251 confirm that this makes sense.


harshilshah4251 (approved):
LGTM

PR #394, opened 2020-10-20
removed relative paths in hs clf python file

harshilshah4251:
removed relative paths in hs clf python file because the script couldnt be executed from outside of the directory otherwise.


cooperka (approved):
LGTM

PR #395, opened 2020-10-20
fix: only send API response if not already sent

karpawich:
Because of the Node.js event loop, API endpoints which utilize the `forEach` Array method to iterate over a list currently try to send multiple responses if multiple errors occur within several iterations of the `forEach` loop.

I've introduced a check of the `res.headersSent` value to each of the affected API endpoints so that only a single response will be sent, even if multiple errors still occur.

**Before**
```javascript
```

**After**

```javascript
if (err) {
    if (!res.headersSent) res.send(err.status, err.message)
        return;
    }
}
```


limmabean (approved):
Bingo that's it.

PR #396, opened 2020-10-20
Conditional for help message for sorting

MrMathur:
I added a function to set $scope.detectHateSpeech to the value in secrets.json (./public/angular/js/controllers/reports/index.js)

I then use this value to determine if batch.html should have the sorting help text through ng-if
(./public/angular/templates/reports/batch.html)


limmabean (approved):
Tried it, works great.

PR #397, opened 2020-10-20
Return of Tag Modal Because Comments are on a Different Branch

limmabean:
I'm returning the tag modal for now because we're developing comments on a different branch and I removed it for comments.


MrMathur (approved):
PR uncommenting previous code. Verified that it outputs expected changes in the front-end.

PR #402, opened 2020-10-23
Pulling Comments Integration UP to Develop

limmabean:
Just yeah/


karpawich (approved):
My review was brief. Looks okay, but did not spend a great deal of time investigating the changes in the Angular HTML that could possibly conflict.

cooperka:
As long as your branch still works that should be fine 👍 We should do a more thorough review when we merge comments into `develop`.

(Note an alternative practice would be to `git rebase` on top of develop before submitting the comments PR, that sometimes comes out a lot more readable. Just advice for the future)

PR #403, opened 2020-10-26
Query performance upgrade

harshilshah4251:
image

Added $text index for author.

Previously, we were doing regex search on all documents for content and authors. This was taking lots of time because regex search is a time-intensive process. 

In this PR, I changed it to two step process. First we find documents matching any of the terms in content or author in the documents using the $text index (which is very fast). In second step, we do regex matches on the returned set (from first step) to find exact matches. This approach takes much less time than before.

### DIFF
--- a/models/query/report-query.js
+++ b/models/query/report-query.js
@@ -51,8 +51,9 @@ ReportQuery.prototype.toMongooseFilter = function() {
   filter = _.omitBy(filter, _.isNil);
   if (this.before)    filter.storedAt = { $lte: this.before }
   if (this.after)     filter.storedAt = Object.assign({}, filter.storedAt, { $gte: this.after });
-  if (this.author)    filter.author = { $regex: this.author, $options: 'i'}
-  if (this.keywords)  filter.content = { $regex: this.keywords,  $options: 'i' }
+  if (this.author || this.keywords) filter.$and = [];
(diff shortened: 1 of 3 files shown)


models/query/report-query.js
```diff
@@ -51,8 +51,11 @@ ReportQuery.prototype.toMongooseFilter = function() {
   filter = _.omitBy(filter, _.isNil);
   if (this.before)    filter.storedAt = { $lte: this.before }
   if (this.after)     filter.storedAt = Object.assign({}, filter.storedAt, { $gte: this.after });
-  if (this.author)    filter.author = { $regex: this.author, $options: 'i'}
-  if (this.keywords)  filter.content = { $regex: this.keywords,  $options: 'i' }
+  //Two step search for content/author. First search for any terms in content or author using the indexed $text search.
+  //Second step is to match exact phrase using regex in the returned superset of the documents from first step.
+  if (this.author || this.keywords) filter.$and = [{$text: { $search: this.author +" "+ this.keywords}}];
```
cooperka: Note: `null + " " + undefined === "null undefined"` so this could unexpectedly narrow results unless we know author and keywords must be empty strings like `""`. To be safer can we do:

```js
{ $search: `${this.author || ""} ${this.keywords || ""}` }
```

cooperka (commented):
Small fix, otherwise looks great

PR #406, opened 2020-10-27
Max/comments integration

limmabean:
I need to fix the tag modal but I'm opening this up now because everything else is fixed.


lib/api/v1/report-controller.js
```diff
@@ -151,10 +162,7 @@ module.exports = function(app, user) {
         // Mark each report as flagged to catch it in model
         report.toggleFlagged(req.body.flagged);
         report.save(function(err) {
-          if (err) {
-            if (!res.headersSent) res.send(err.status, err.message)
-            return;
-          }
+          if (err) return res.send(err.status, err.message);
```
cooperka: Are these all reverted because it's handled automatically now in https://github.com/TID-Lab/aggie/pull/398 or is that separate?

config/secrets.json.example
```diff
@@ -15,6 +15,12 @@
     "useLanguage": false,
     "zawgyiProb": 0.9
   },
+  "comments": {
+    "username": "Enter username",
+    "password": "Enter password",
+    "baseUrl": "Enter base URL",
+    "count": 100
```
cooperka: `count` is not very descriptive. It looks like it's only used a few places, could we be more verbose? (I'm not actually sure what it does after reading the code, since it's just sent as params to the API)

lib/fetching/content-services/comments-content-service.js
```diff
@@ -0,0 +1,170 @@
...
+
+        if (comments.length === 0) {
+            return callback([]);
+        }
+
+        this.cache = {};
+
+        const { acquiredAt, timestamp } = comments[comments.length - 1];
+
+        // TODO rewrite with async/await to avoid callback hell
```
cooperka: 😁 

(It's pretty straightforward to simplify this if you wanna take a swing at it, happy to send a quick example I have. Not blocking, just nice.)

lib/fetching/content-services/comments-content-service.js
```diff
@@ -0,0 +1,170 @@
...
+
+    const report = {
+        authoredAt: new Date(doc.timestamp),
+        fetchedAt: new Date(),
+        content,
+        url,
+        metadata,
+    }
+
+    const post = this.cache[postUrl];
```
cooperka, line 180: This memoization can be great for performance, but do you know if it's really necessary? Generally `findOne` should be very fast assuming reports are indexed on `url`. I worry this might be premature optimization that adds complexity without explicit benefit.

models/report.js
```diff
@@ -85,11 +88,23 @@ schema.methods.addSMTCTag = function(smtcTagId) {
   });
   if (isRepeat === false) {
     this.smtcTags.push({_id: smtcTagId});
+
+    if (!this.commentTo && this._media[0] === 'crowdtangle' && !this.url.match(/permalink/)) {
```
cooperka, line 93: This is a lot of conditions; looks fine but seems ripe for a nice "why" comment in the code.

public/angular/js/controllers/reports/show.js
```diff
@@ -1,141 +1,205 @@
 angular.module('Aggie')
 
-.controller('ReportsShowController', [
+  .controller('ReportsShowController', [
```
cooperka: Your IDE may have done this automatically, but everything else is unindented here so please undo this to keep things consistent.

public/angular/js/controllers/reports/show.js
```diff
@@ -1,141 +1,205 @@
...
+      };
+
+      /**
+       * Saves a front-end report to the back end.
+       * @param {Report} report
+       */
+      $scope.saveReport = function(report) {
+        Report.save({ id: report._id }, report, function() {
+        }, function() {
+          flash.setAlertNow("Sorry, but that report couldn't be saved.");
```
cooperka: Would be lovely to at least log an error to the console, lest it be swallowed by the void.

cooperka (commented):
Just some general code notes, I haven't tested this locally or looked at the HTML at all. Thank you for all this work!

lib/api/v1/report-controller.js
karpawich: Separate. They were accidentally reverted, but I just reverted them back.

config/secrets.json.example
karpawich: Yes. The ambiguous name is used in an effort to mimic the CrowdTangle content service so that the comments content service can be edited by Harshil, Daniel, or anyone else who helped to write the CT content service.

I've refactored the name to `pageCount` to reflect its role in telling the API the # of comments to return per page. However, I still want to keep the nomenclature similar so as to ensure others can make edits across both content services.

lib/fetching/content-services/comments-content-service.js
karpawich: I am hesitant to remove the cache, for the reason that I think it will reduce what might be 100 `Report.findOne` calls to 1-3 on average. To explain further:

Assuming the comment acquisition system runs at full capacity, then `_parse` will be called 100 times every ten seconds for each of the 100 comments fetched from the API. On average, these 100 comments will be associated with just 1-3 posts. Thus, caching post reports fetched from the database will reduce 100 separate and mostly redundant indexed `Report.findOne` calls to 1-3.

@cooperka knowing the context better, is `Report.findOne` fast enough when indexed by URL such that even a search by URL through the hundreds of thousands of reports is trivial to execute? If so, I can happily remove this optimization.

config/secrets.json.example
cooperka: Thanks! As long as someone a year from now can look and figure out what it means. pageCount sounds reasonable.

models/report.js
karpawich: Comment added!

lib/fetching/content-services/comments-content-service.js
cooperka: Finding via index should generally be `O(1)` (extremely fast) in all cases. **Searching** (as in, arbitrary regex match) is slower but I don't think you're doing that here.

Anyway, I assume it's currently working so let's keep as is. Don't want to throw a wrench in things this close to the election!

lib/fetching/content-services/comments-content-service.js
karpawich: Code rewritten to use async/await :)

lib/fetching/content-services/comments-content-service.js
karpawich: I'll at least at a TODO note for the future. Did not know that indexed searches are O(1)!

PR #409, opened 2020-10-28
Tag filter

RoboticWater:
Added a search bar to the tag sidebar which lets users filter the tag box. It works by hiding tags that don't contain the substring in the search bar, so tags won't have consistent locations, but will have a consistent order (e.g. filtering 'a' in tags a, b, aa will always result in a, aa). This may not be ideal for consistency, but it's better to have the filter functionality.


RoboticWater:
Ugh, I keep forgetting to reset the CT list JSON. >:/

cooperka:
Tip: `git update-index --skip-worktree config/crowdtangle_list.json`

limmabean (approved):
You're going to have some merge conflicts I think. Make sure the tags have the x-button.

PR #410, opened 2020-10-28
excluding gateway timeout

harshilshah4251:
not sure if this is the best way to address it. I tested it using the Unauthorized error code 401. Gateway timeout is hard to reproduce because it depends on when Crowdtangle API doesnt respond to Aggie's request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,123 @@ sudo $EDITOR /etc/logrotate.conf
 # Whenever you need to, you can view app logs by running (in the `aggie` directory):
 npx pm2 logs
 ```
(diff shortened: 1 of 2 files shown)


cooperka (approved):
Fix LGTM!

There were other changes on this branch so instead of merging this PR I cherry-picked your commit onto develop (see bfc36fd31d2bc4da78fa1d3d7c2598a4239e2c86).

Thanks, and if it happens again we'll have more detailed logs to figure out what type of response it returns (it may not be 503 in this case; perhaps the promise gets killed in some other way when it takes too long to return. No way to know without trying.)

PR #411, opened 2020-10-29
Added date filter to incident page

RoboticWater:
Adds a date filter to the Incident tab that functions identically to one on the Report page.

Note: IncidentQuery is never used outside of tests, so we should either get rid of it, or rewrite the incident controller to use it. For the sake of consistency with the report controller, we should probably use it.


limmabean (approved):
Tested, went through code. Looks great.

PR #413, opened 2020-11-02
Make clicking the "___ new reports" flash refresh the page

limmabean:
This is a rather rudimentary fix for a tag issue (adding tags to new reports breaks any future ability to add tags to any report) but prevents end-users from breaking the system unintentionally. I left a TODO in there to make the original method work but, this is not outside the realm of how systems behave. When there are new changes to the reports, the page will request you refresh the page.

### DIFF
--- a/public/angular/js/controllers/reports/index.js
+++ b/public/angular/js/controllers/reports/index.js
@@ -4,6 +4,7 @@ angular.module('Aggie')
   '$scope',
   '$rootScope',
   '$stateParams',
+  '$window',
   'FlashService',
   'reports',
   'sources',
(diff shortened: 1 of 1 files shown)


public/angular/js/controllers/reports/index.js
```diff
@@ -505,10 +509,13 @@ angular.module('Aggie')
      * @param {SMTCTag} smtcTag
      */
     $scope.addTagToSelected = function(smtcTag) {
+      console.log(smtcTag);
       //TODO: There should be a validation that the tag is not already added to the report
       var items = $scope.filterSelected($scope.reports);
       if (!items.length) return;
       var ids = getIds(addSMTCTag(items, smtcTag));
+      console.log(ids);
+      console.log(smtcTag._id);
```
cooperka, line 518: Can we remove these?

cooperka (approved):
Reasonable workaround, thanks for the clear documentation and explanation.

Suggestion to remove logs (unless you want them for debugging in production) then feel free to merge.

RoboticWater (approved):

PR #415, opened 2020-11-04
Mataomo Breaks the Front-end If the config settings isn't available.

limmabean:
The lack of the setting should either prevent the system from working initially or not prevent the front-end from working. I've updated the secrets.json.example and made it so that the front-end still works even if the config was not set.


MrMathur (approved):

cooperka:
Thanks! I changed it to a regular log because many servers just won't have this set up at all.

PR #416, opened 2020-11-04
Quick & Dirty Tag Filter Fix

RoboticWater:
This doesn't fix the underlying CSS issue (i.e. the CSS doesn't seem to be compiling on production), which is definitely still a problem, but for the sake of the election, this should hopefully suffice.

### DIFF
--- a/public/angular/sass/components/tags.scss
+++ b/public/angular/sass/components/tags.scss
@@ -8,21 +8,6 @@ tags.input--tags {
   --tags-hover-border-color: #CCC;
 }
 
-
-#tag-search {
-  width: 100%;
-  border: 1px solid #0002;
-  padding: 5px 6px;
-  font-size: 1.1em;
-  border-radius: 3px;
-  background: #fff;
-}
-
-#tag::placeholder {
-  color: #f00;
-}
-
-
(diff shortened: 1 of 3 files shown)


limmabean (approved):
Tested on the Aggie Demo Server. Works great.

cooperka:
If I understand correctly, this is a workaround to get some specific CSS to update because it's not updating the proper way? Let's focus on figuring out where things went wrong and fix it there instead. `git bisect` will help a lot here, we just need to figure out what command generates CSS (in order to test it quickly) and why it's different in production. Perhaps `NODE_ENV=production` is the setting that affects this behavior.

I don't think we should merge this PR (but it's a clever solution 😉). Solutions like this only work for servers that were previously on a very specific commit and thus only have this one change to their CSS. Until we fix the root problem, you can generate CSS locally and `scp ./public/angular/css/app.css azureuser@1.2.3.4:aggie/public/angular/css/` to get it on the server 💩 I did this successfully for a new server that Michael created this afternoon.

cooperka:
CSS overwritten now FYI:
 
Screen Shot 2020-11-04 at 18 16 22

PR #417, opened 2020-11-04
Added Hate Speech Icon based on Threshold defined in Config

limmabean:
Could someone with the hate speech algorithm running tell me if this works?

### DIFF
--- a/config/secrets.json.example
+++ b/config/secrets.json.example
@@ -112,7 +112,7 @@
   "adminEmail": "aggie-admin@example.com",
   "detectHateSpeech": false,
   "hateSpeechThreshold": {
-      "threshold": 60,
+      "threshold": 20e-2,
       "enabled": true
     },
 }
--- a/config/secrets.json.example
+++ b/config/secrets.json.example
@@ -102,7 +102,6 @@
       "filename": "logs/analytics.log"
     }
   },
-
   "api_request_timeout": 60,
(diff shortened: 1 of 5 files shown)


harshilshah4251 (commented):
Tested and works great.
image

config/secrets.json.example
```diff
@@ -102,12 +102,17 @@
...
   "secret": "Replace this string when installing aggie with something else!!",
   "adminParty": false,
   "experiment": false,
   "experimentFile": "test/end-to-end/fixtures/experiment_reports.json",
   "adminPassword": "letmein1",
   "adminEmail": "aggie-admin@example.com",
-  "detectHateSpeech": false
+  "detectHateSpeech": false,
+  "hateSpeechThreshold": {
+      "threshold": 20e-2,
```
harshilshah4251, line 114: maybe add a comment there that this value needs to be between 0 and 1

cooperka: Agree this should be better documented, unfortunately JSON doesn't allow comments... the readme would be a good spot for this.

Is 20% thought to be a sensible default? Most server admins will accept it blindly.

config/secrets.json.example
```diff
@@ -102,12 +102,17 @@
       "filename": "logs/analytics.log"
     }
   },
+
```
cooperka: Please remove

cooperka (approved):
Besides documentation, code looks great.

PR #419, opened 2020-11-05
changing default fetching frequency and adding index for dates

harshilshah4251:
increasing freq interval to avoid "Too many requests errors" after restart and adding index for dates.


limmabean (approved):
Perfect. Checked the index and did a date search.

PR #420, opened 2020-11-06
Dealing with null values for batch sorting

MrMathur:
Created a new property in metadata of visible reports that stores hateSpeechScore, and assigns a score of -1 for null values. tableBatch.html now sort with the new property (sortingScore)


harshilshah4251 (approved):
LGTM

harshilshah4251:
image

PR #422, opened 2020-11-07
Fetching Login Id for aggie data

MrMathur:
Screenshot 2020-11-06 at 11 17 27 PM

### DIFF
--- a/public/angular/js/controllers/application.js
+++ b/public/angular/js/controllers/application.js
@@ -28,34 +28,37 @@ angular.module('Aggie')
         var u='https://' + data.matomo.dashboard_name + '.matomo.cloud/';
         _paq.push(['setTrackerUrl', u+'matomo.php']);
         _paq.push(['setSiteId', data.matomo.site_id]);
-        var d=document, 
-          g=d.createElement('script'), 
+        var d=document,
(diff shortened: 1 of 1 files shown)


public/angular/js/controllers/application.js
```diff
@@ -47,6 +48,14 @@ angular.module('Aggie')
       g.src='https://cdn.matomo.cloud/' + data.matomo.dashboard_name +'.matomo.cloud/container_' + data.matomo.container_id  + '.js'; 
       s.parentNode.insertBefore(g,s);
 
+      // Sending Username Data
+      $rootScope.$watch('currentUser', function() {
+        if ($rootScope.currentUser != undefined) {
+          var user = $rootScope.currentUser;
+          _paq.push(["setUserId", user.username]);
+        }
```
cooperka: Have you tested this with different usernames -- what happens if you sign out and sign back in as someone else? Just want to make sure it properly attributes in that case.

Let's also add an `else` condition to push `["setUserId", null]`, that seems simple enough right?

UPDATE: Fixed.

cooperka (commented):
Exciting that you got this working in the same file as everything else! That keeps everything nice and consolidated. Just one comment.

PR #426, opened 2020-11-08
Visualization CT Source title fix

RoboticWater:
Truncates CT Source titles that are longer than 25 characters so they don't overlap, and removed unnecessary console log


cooperka (approved):

cooperka:
Still some overlap 😉 but definitely improved.

Screen Shot 2020-11-07 at 22 30 34

RoboticWater:
Yeah, I'll need to figure out how to make it dynamic. I think there's a way, so it might be ready by Tuesday.

PR #428, opened 2020-11-24
Analysis Fixes: Graphs now scale + render time in Myanmar/Yangon time zone

RoboticWater:
This should fix graphs going off the screen, and the time zones being based on local user time. Note that I've added `moment-timezone` to the package.json, so you'll need to `npm i` for this one.

### DIFF
--- a/public/angular/js/controllers/analysis/alg-hate-speech-graph.js
+++ b/public/angular/js/controllers/analysis/alg-hate-speech-graph.js
@@ -1,3 +1,5 @@
+var responsivefy = require('./chart-utils');
+var moment = require('moment-timezone');
 module.exports = function renderHateSpeechReportGraph(id, data, range, description) {
   var margin = { top: 50, right: 0, bottom: 55, left: 50 };
   var width = 1200,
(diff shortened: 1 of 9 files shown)


public/angular/js/controllers/analysis/alg-hate-speech-graph.js
```diff
@@ -13,14 +15,14 @@ module.exports = function renderHateSpeechReportGraph(id, data, range, descripti
           d.value +
           " Reports" +
           "</div><div>" +
-          d3.timeFormat("%b. %d %H:00")(d.date).replace(/\s0/g, " ") +
+          moment(d.data.date).tz("Asia/Yangon").format('MMM. Do hh:mm:ss') +
```
harshilshah4251, line 18: I should have said this earlier but I think since the Myanmar project is over, do we still want to make this change? We can maybe just make this change on the Myanmar server and not aggie repo. Lmk if I am missing somthing.

harshilshah4251: LGTM

PR #429, opened 2020-11-24
Fixed the following

limmabean:
-Make batch work with tag search (though kinda dumb)
-Made it so that trackers can click an external link right on the reports or batch page.
-Fixed text overflow on the incidents show page.
-Returned tags to put model but didn't add back the buttons to tag modals.

### DIFF
--- a/public/angular/templates/incidents/show.html
+++ b/public/angular/templates/incidents/show.html
@@ -23,41 +23,43 @@
           </div>
         </div>
         <div class="card-body">
-          <table class="table table-bordered">
+          <table class="table table-bordered incidentShow__table">
             <tbody>
-            <tr>
-              <th class="detail-key" translate>Location</th>
-              <td class="detail-value">
-                <span ng-if="incident.locationName">{{ incident.locationName }}</span>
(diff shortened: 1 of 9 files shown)


harshilshah4251:
When I type in tag and click on search, it automatically shows me the tag id instead of the tag name. If you can fix this, it would be great.
image

- Also, I don't know how to test the tag batch functionality. I searched for a tag and then clicked on grab batch but doesn't work for me. If you can make sure that it works as expected, it would be great. 

Everything else works great.

public/angular/templates/incidents/show.html
```diff
@@ -23,41 +23,43 @@
           </div>
         </div>
         <div class="card-body">
-          <table class="table table-bordered">
+          <table class="table table-bordered incidentShow__table">
```
harshilshah4251, line 26: is this a naming convention to have __ (double underscores)?

harshilshah4251:
I also added 2 line change for filter by author date since it was easy. You can review that and merge it with this PR.

public/angular/templates/incidents/show.html
limmabean: Yeah its BEM

PR #436, opened 2021-01-31
Remove order by status

MrMathur:
**Description:** Incidents should be ordered in pagination by their ID number, with the highest number on the earliest page. Instead there is some sort of priority given to open incidents by default. This is nonsensical and makes it difficult for users to find the incident they want by ID.

Removed the ordering by status for the incident table. The change can be observed by going to the incident page, and having a bunch of incidents that are a mix of open and closed. The closed incidents do not migrate to the bottom of the list anymore.


limmabean (approved):
Tested, works great.

PR #437, opened 2021-02-01
Johnny dev 1

limmabean:


limmabean:
Try not to change the package-lock.json. It doesn't really matter but its annoying to resolve merge conflicts with something that doesn't exist until runtime.

PR #439, opened 2021-02-02
Incident Search Updates

ciabhanconnelly:
Added the ability to search for a specific incident ID, and to filter by author with a dropdown.


limmabean:
image

Just removes the incidents I have without me searching anything.

limmabean:
image
It doesn't break my incidents now, but it does not work.

limmabean:
image
Whenever I get a problem where I don't see any incidents on the incident page on default, I can search by ID but not by user.

limmabean (approved):
Works like a charm now.

PR #440, opened 2021-02-05
Twitter config info

akjadhav:
Updated on twitter config popup...

- consumer_key and consumer_secret to api_key and api_secret
- Link to app.twitter.com to developer.twitter.com


akjadhav:
Just double checked with another Twitter key and it doesn't seem to work.

limmabean (approved):
LGTM

PR #441, opened 2021-02-08
Report level escalation 

limmabean:
NOT FINISHED DO NOT MERGE.

I have finished the functionality but need to test the system before merging. Can someone just check if the icons change through refreshes on the front-end.


limmabean:
Will commit with Relevant Reports

PR #442, opened 2021-02-08
Report veracity backend

sairapoonnen:
Adding API endpoint for veracity


limmabean (commented):
On line 144 of report-controller, you wrote req.body.flagged. Change this to a veracity value. Also veracity has 3 values: Confirmed True, Confirmed False, and Uncomfirmed. Its not a boolean value.

limmabean (approved):

PR #445, opened 2021-03-01
Fixed Bug

ciabhanconnelly:
Fixed the pagination bug with a quick fix within the problem function to handle when it is passed a string rather than an array of string. However, it might be good to eventually hunt down where $scope.searchParams.tags is not being put into a list when a single tag is searched and not on the first page.


limmabean (approved):
Looks Great

PR #446, opened 2021-03-01
Frontend flag deprecation

akjadhav:
Flag options removed from reports page and reports batch page.

Users will reports currently flagged will not be affected (the reports will just remain flagged and there's no more options to flag/unflag new ones).


akjadhav:
This has been checked on the frontend and Aggie works as usual. Not tested with the backend deprecation yet.

limmabean:
You do need to remove a lot more than this. The front-end controllers have references to Flags. Schema has references to flags. Both front-end and backend API endpoints have flag code.

PR #448, opened 2021-03-02
Remove flags backend and frontend

cuongnguyenx:
Tested on my machine, and everything seems to work normal for both the front and backend.


akjadhav:
Note that the flag deprecation for the Aggie docs is in a separate PR

PR #454, opened 2021-03-15
Done

ciabhanconnelly:
I think this should work. Having issues pulling from CT on my local machine, as is Johnny, so will need someone to test this. Our issues are unrelated to/persist without this modification.


karpawich (changes requested):
It looks as though the only change needed for the `secrets.json.example` file is the introduction of the interval key-value pair to the CrowdTangle config object.

karpawich (approved):

limmabean (approved):
LGTM

karpawich (approved):

PR #461, opened 2021-03-29
Incident tags

limmabean:
Currently there is an issue with not seeing new Tags, but this is also because incidents don't update when they are changed, we should change this. There is no socket streaming for incident changes, only incident creation.


limmabean:
Merged with Relevant Reports Page

PR #466, opened 2021-04-04
Dashboard vis

MrMathur:
Reading from Mongo collection created through analysis process, and creating visualisation front-end using that data.


karpawich (approved):
lgtm!

I made a few simple modifications. See the commits for details

PR #467, opened 2021-04-05
Vis dashboard update

MrMathur:
Fixes errors when the analysis visualisations are passed empty data. Also fixes the timeline chart date parsing erros.


karpawich (approved):
lgtm (upon brief inspection!)

PR #468, opened 2021-04-05
Sorting groups by clicking table header

MrMathur:
Sorting groups by clicking on the group view table headers.


limmabean:
I merged this with the quick fix branch.

PR #481, opened 2021-04-07
Selection validation

akjadhav:
Added user alerts if action is attempted without selecting required groups/reports


limmabean (approved):
LGTM

PR #486, opened 2021-04-08
Filtering interaction

MrMathur:
Click on tags to filter all views by data from specific tags


MrMathur:
Should be merged after [https://github.com/TID-Lab/aggie/pull/485](https://github.com/TID-Lab/aggie/pull/485) has been merged

limmabean (approved):
I can confirm that this won't break the rest of aggie. But the visualization isn't really working on my computer.

PR #548, opened 2024-09-03
Batch mode

codelastnight:
- add batch mode to reports page
- add useReportsMutations for reusable useMutations 

some other stuff:
- Add AggieCheck (reusable checkmark styled)
- remove more Bootstrap references from pages.
- move ReportListItemSmall into SocialMediaPost
- begin to refactor types (types from backend should originate from /api, by doing so we have 1 (folder) source of truth
- deprecate useOptimisticMutation (i didn't consider how complex optimistic mutations can get...)


codelastnight:
i opened pull request in the wrong repo....

PR #549, opened 2024-09-10
Comments

codelastnight:
- add comment suppport to group documents
- design spec for individual incident 
- minor bugfixes with tanstack query
- edit incident from inside incident page
- periodically update from backend


codelastnight:
i did it again 💀

